PR #4, opened 2016-02-12
Instances

adamnfish:
Adds instance and instance AMIs to Amiable. Still not displayed nicely and no UI for getting to them.


philmcmahon:
:+1:

PR #6, opened 2016-02-15
Add date colours

philmcmahon:
Makes dates pretty print, with the number of days since an AMI was created highlighted in red/amber/green.

Could do with some refactoring to avoid the repeated template code


adamnfish:
I've used the changes on this branch in another PR (https://github.com/guardian/amiable/pull/8) since it doesn't build.

PR #7, opened 2016-02-17
Optional Stack Stage App filters

adamnfish:
Allows users to see instances by Stack/Stage/App without having to fill in all three.

Still loads of issues and template stuff to sort.


sihil:
Beautiful. :+1:

PR #8, opened 2016-02-18
Template improvements

adamnfish:
Pulls in template changes from https://github.com/guardian/amiable/pull/6 and improves layout.


app/models/models.scala
```diff
@@ -69,3 +69,8 @@ case class AMIableError(
   statusCode: Int,
   context: Option[String] = None
 )
+
+trait Age
```
sihil: sealed?

adamnfish: Oops

sihil:
:+1:

PR #9, opened 2016-02-18
Do not choke on missing amis

adamnfish:
Instance AMIs will now show failed lookups in among the successes rather than completely dieing the first time something fails. This is *much* more useful.

The sorting has been removed since it was no longer trivial and this will completely change anyway. The aim is to show these AMIs by stack, stage and app.

(Based on https://github.com/guardian/amiable/pull/8 so that must be merged first)


sihil:
:+1:

PR #10, opened 2016-02-19
Design overhaul

adamnfish:
Now focuses on what AMIable actually does, with placeholders for upcoming features.

Features dramatically reduced eye-pain.


sihil:
Go for it. :+1:

PR #11, opened 2016-02-22
Add analytics to homepage

adamnfish:
Generate more useful insights from the data by working out the number of affected PROD instances and the stacks where they live.

This implements the TODOs on the homepage as well as a couple of other teeny tidies that change touched.


sihil:
:+1:

PR #17, opened 2016-02-29
More helpful ami display

adamnfish:
Displays instance AMIs by stack/app/stage to show their usage.

Note: This change drops support for showing AMIs that could not be looked up. If that's required it'll need to be added in another PR. One change at a time!

ssa-amis

### DIFF
--- a/test/util/Fixtures.scala
+++ b/test/util/Fixtures.scala
@@ -1,8 +1,16 @@
 package util
 
-import play.api.libs.json.{JsObject, JsValue}
+import models.{AMI, Instance, Meta, Origin}
+import org.joda.time.DateTime
 
 object Fixtures {
+  def emptyAmi: AMI =
+    AMI("", None, "", "", None, Map.empty, None, "", "", "", "", "",  None)
+  def emptyInstance(arn: String): Instance =
+    Instance(arn, "", "", "", "", "", DateTime.now, "", "", "", Nil, Map.empty, None, None, Nil, Nil, Map.empty, Meta("", Origin("", "", "", "")))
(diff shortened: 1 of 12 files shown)


test/util/Fixtures.scala
```diff
@@ -1,8 +1,18 @@
...
+import models._
+import org.joda.time.DateTime
 
 object Fixtures {
+  def emptyAmi(arn: String): AMI =
+    AMI(arn, None, "", "", None, Map.empty, None, "", "", "", "", "",  None)
+  def emptyInstance(arn: String): Instance =
+    Instance(arn, "", "", "", "", "", DateTime.now, "", "", "", Nil, Map.empty, None, None, Nil, Nil, Map.empty, Meta("", Origin("", "", "", "")))
+  def instanceWithAmiArn(arn: String, amiArnOpt: Option[String]): Instance =
+    amiArnOpt.fold(emptyInstance(arn))(amiArn => emptyInstance(arn).copy(specification = Map("imageArn" -> amiArn)))
```
sihil: Ha. Somehow I'd never cracked the functional way to do this. Nice pattern.

sihil:
In my mind this would be the other way round and show an AMI with the list of SSAs that are associated with it. Also I think that we should adopt a concise format for the SSA, something like stack:app/stage.

adamnfish:
Soz, merged this for the demo. I misagree because you search by SSA, not by AMI. Let's chat IRL though.

sihil:
Yeah - I'm not particularly attached to it and I think this makes sense. Although I still think we should have a consistent SSA format.

PR #18, opened 2016-03-02
Improve SSA display on instance AMIs page

adamnfish:
Lists all found AMIs first, then shows usage. Consistently displays the SSA in a table for clarity.

instancessa-display


adamnfish:
@sihil As you said in the last PR, it wasn't obvious exactly what you were looking at. I also spotted that some accounts (e.g. Frontend) had millions of SSA combos but only a couple of distinct AMIs. We show the distinct AMIs first to make this much clearer.

sihil:
:+1:

PR #19, opened 2016-03-15
Update to Play 2.5

adamnfish:
Play application state (WS / config) are now injected into the Controller.

This is to update to the latest version of Play before I start adding lifecycle-based components to the application (which would make it much harder to migrate Play version).

### DIFF
--- a/app/controllers/AMIable.scala
+++ b/app/controllers/AMIable.scala
@@ -1,21 +1,20 @@
 package controllers
 
+import javax.inject.Inject
+
 import config.AMIableConfig
-import models.{SSA, Attempt}
+import models.{Attempt, SSA}
 import play.api._
 import play.api.libs.ws._
 import play.api.mvc._
-import prism.Prism
-import prism.PrismLogic
+import prism.{Prism, PrismLogic}
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext
 
 
-class AMIable extends Controller {
+class AMIable @Inject()(ws: WSClient, playConfig: Configuration)(implicit exec: ExecutionContext) extends Controller {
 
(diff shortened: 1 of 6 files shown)


app/controllers/Healthcheck.scala
```diff
@@ -1,8 +1,10 @@
 package controllers
 
+import javax.inject.Inject
+
 import play.api.mvc._
 
-class Healthcheck extends Controller {
+class Healthcheck @Inject() extends Controller {
```
sihil: Ugg. So damn unpretty.

sihil:
:+1: Also - shall we fork Play already and lose the DI?

app/controllers/Healthcheck.scala
adamnfish: I know!!!
To be fair, the other controller comes out looking a bit neater, although at the cost of it using impenetrable Dark Magiks.

app/controllers/AMIable.scala
```diff
@@ -1,21 +1,20 @@
...
-import prism.Prism
-import prism.PrismLogic
+import prism.{Prism, PrismLogic}
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext
 
 
-class AMIable extends Controller {
+class AMIable @Inject()(ws: WSClient, playConfig: Configuration)(implicit exec: ExecutionContext) extends Controller {
```
cb372: :mask:

adamnfish: I may switch to the compiled version at some point, but I just wanted to upgrade to Play 2.5 with as small a change as possible. Sorry.
As we've discussed before though, I'm deeply unsatisfied with both approaches so I'm still on the fence about moving off the "simpler" (sic) and better documented default version.

PR #20, opened 2016-04-06
Recommend amis

adamnfish:
Improves display and adds upgrade recommendations to AMIs.

When an AMI is out of date, that's clearly displayed and AMIable tries to work out an AMI that should be used instead.
upgrade-available

When an AMI is up to date it is happily green and AMIable decides that there is nothing to upgrade to.
up-to-date

The PR also removes a bunch of redundant functionality. Sorry it's quite a big one, this was drips of work over many weeks.


test/prism/RecommendationsTest.scala
```diff
@@ -0,0 +1,64 @@
+package prism
```
philmcmahon: needs a few more tests for the recommendations :trollface:

philmcmahon:
:+

test/prism/RecommendationsTest.scala
adamnfish: Great spot, very true. I promise to add them later.

Issue #14, opened 2016-02-23
Store statistics in cloudwatch


sihil:
In order to analyse the trends (such as how awesome an effect AMIable had on the recency of AMIs) it would be great to store data on a regular basis of the number of old AMIs in use let's store stats in cloudwatch.

adamnfish:
Stats are going into cloudwatch and graphed on the homepage.

PR #21, opened 2016-04-08
Add agent to provide old instance count to cloudwatch

adamnfish:
Sends "the number of PROD instances running out of date AMIs" to cloudwatch as a metric. This is something that can be used to track the effectiveness of AMIable.


adamnfish:
Fixes https://github.com/guardian/amiable/issues/14

sihil:
:+1:

PR #22, opened 2016-04-11
Display old instance count history

adamnfish:
Adds a graph of old PROD AMI count to the index page now that data is being collected.

graph


sihil:
👍

PR #24, opened 2016-04-14
Side by side upgrade recommendation

adamnfish:
UI changes to bring upgrade information into view. It now displays the correct upgrade alongside the AMI in the results page and the AMI info page. This makes it easy peasy to work out which AMI to upgrade to and verify it is correct.

### Side by side upgrade

side-by-side-upgrades-1

### No upgrade available

side-by-side-upgrades-2

### Unable to work out upgrade

side-by-side-upgrades-3

### Upgrade icons on list display

upgrade-icons


sihil:
All sorts of awesome in a single PR. 👍

PR #25, opened 2016-04-15
Trigger modal by clicking anywhere on AMI badge

adamnfish:
Makes the entire AMI badge clickable. Especially important now that the styling no longer contains a "link".


adamnfish:
Looks like there's a bug in MaterializeCSS that causes problems with programmatically opened modals.

PR #31, opened 2016-06-02
Adds play-googleauth to Amiable

adamnfish:
DO NOT MERGE (yet)

Adds the required Google configuration properties and code. Includes the play googleauth library as a dependency (dev version see below) to provide Google auth functionality. It also adds the controller and required routing / templates to support login and logout behaviours.

For dev mode, this includes an nginx setup script. The DEV config file now needs to include Google credentials to run.

**Note**:
This uses the WIP version of play-googleauth. It serves to test that pull request (https://github.com/guardian/play-googleauth/pull/39). This should be merged only after that is released, and with a proper published version of that library as a dependency. This also means it won't build, since the snapshot dependency isn't available outside of my local ivy cache.

It will also require a PROD config change before release.


adamnfish:
cc @emma-p I used this to test the Play-GoogleAuth integration. Seems to work ok!

emma-p:
@adamnfish Great, 👍 to merge it then once the other pull request is merged

PR #32, opened 2016-09-16
Use a .deb for deployment

philwills:
I wanted to get off Trusty and since that involved getting onto SystemD, thought I'd move over to `deb` packaging at the same time.

https://github.com/guardian/deploy-tools-platform/pull/43 needs to happen in parallel.


build.sbt
```diff
@@ -34,4 +23,40 @@ resolvers += "scalaz-bintray" at "http://dl.bintray.com/scalaz/releases"
...
+maintainer := "Guardian Developers <dig.dev.software@theguardian.com>"
+packageSummary := "AMIable"
+packageDescription := """Web app for monitoring the use of AMIs"""
+debianPackageDependencies := Seq("openjdk-8-jre-headless")
+
+javaOptions in Universal ++= Seq(
+  "-Dpidfile.path=/dev/null",
+  s"-Dconfig.file=/etc/${name.value}.conf",
+  "-J-XX:MaxRAMFraction=2",
+  "-J-XX:InitialRAMFraction=2",
```
philwills: This does mean 1/2 of RAM as opposed to 2/3, but that doesn't seem a big problem.

build.sbt
```diff
@@ -34,4 +23,40 @@ resolvers += "scalaz-bintray" at "http://dl.bintray.com/scalaz/releases"
 
 javaOptions in Test += "-Dconfig.file=conf/application.test.conf"
 
-playDefaultPort := 9101
+PlayKeys.playDefaultPort := 80
```
adamnfish: Should stay the same

build.sbt
adamnfish: Yeah, it won't matter here. Be good to know how to set it to arbitrary fractions for stuff that will care more. 512 -> 768 is a big difference on a micro and 256 -> 384 might be the difference between working and not working on a nano. (I assume it's fairly easy)

build.sbt
```diff
@@ -1,19 +1,8 @@
-import com.gu.riffraff.artifact.RiffRaffArtifact
-import com.gu.riffraff.artifact.RiffRaffArtifact.autoImport._
-import play.sbt.PlayImport.PlayKeys._
-
-
 name := """amiable"""
 
 version := "1.0-SNAPSHOT"
 
-lazy val root = (project in file("."))
```
adamnfish: I'm confused. Where's the project definition now?

adamnfish: Oh, I guess you're just using whole the file as a single project? I prefer to explicitly write the project every time because that approach works in every sbt project, rather than using different approaches for single and multi project builds.

build.sbt
```diff
@@ -34,4 +23,40 @@ resolvers += "scalaz-bintray" at "http://dl.bintray.com/scalaz/releases"
...
+def env(key: String): Option[String] = Option(System.getenv(key))
+
+riffRaffBuildIdentifier := env("BUILD_NUMBER").getOrElse("DEV")
+riffRaffManifestBranch := env("BRANCH_NAME").getOrElse("unknown_branch")
+riffRaffManifestVcsUrl := "git@github.com:guardian/amiable.git"
+riffRaffUploadArtifactBucket := Some("riffraff-artifact")
+riffRaffUploadManifestBucket := Some("riffraff-builds")
+riffRaffArtifactResources  := Seq(
+  riffRaffPackageType.value -> s"packages/${name.value}/${name.value}.deb",
+  baseDirectory.value / "deploy.json" -> "deploy.json"
```
adamnfish: @sihil says this isn't required (it would discover the deploy.json)?

philwills: That's only true if you rely on the defaults, but we're renaming the `deb`, as the default is to  include a version number and architecture (all) in the filename, which is a bit messy. Having said that, I could definitely be convinced that putting that cruft in the userdata is better than adding four lines of noise here.

build.sbt
philwills: I just see it as noise for single projects, but I don't feel strongly about it.

build.sbt
adamnfish: I meant the `deploy.json` line, rather than the `deb` bit. I was chatting with Simon about having the plugin be able to work out these bits so this noise isn't required. This is ok for now though.

build.sbt
adamnfish: Meh. I prefer the other way but not enough to change it now ;-)

adamnfish:
Thanks Phil!

:+1:

adamnfish:
This branch deployed fine with the new CF template applied, so I'll merge this now.

PR #33, opened 2017-02-02
Show instance total count next to out-of-date count

TBonnin:
Out-of-date instances info is more meaningful when compared to the total number of instances

Before:
screen shot 2017-02-02 at 10 31 40

After:
screen shot 2017-02-02 at 10 31 28


adamnfish:
:+1: Fantastic idea and it looks great.

PR #35, opened 2017-02-16
Add Instance AMIs age percentile and old instances count on both index and instanceAMIs page

TBonnin:
Global age percentiles are not stored yet. I will do that in a subsequent PR so we can track them over time. 

screen shot 2017-02-21 at 13 07 25

screen shot 2017-02-21 at 13 07 11


app/views/instanceAMIs.scala.html
```diff
@@ -1,28 +1,32 @@
-@(ssa: SSA, amisWithUpgrades: List[AMI], amiSSAs: List[(SSA, scala.List[AMI])])
+@(ssa: SSA, amisWithInstances: List[(AMI, List[Instance])], amiSSAs: List[(SSA, List[AMI])], agePercentiles: utils.Percentiles)
```
TBonnin: We should probably have a `case class AmiWithInstances(ami: AMI, instances: List[Instance])` rather than a tuple `(AMI, List[Instance])` so we can avoid dealing with the awkward `._1` below. What do you think?

TBonnin: I have done it there d0d562a

app/prism/Prism.scala
```diff
@@ -37,11 +38,23 @@ object Prism {
...
+  private def instancesAndAmis(stackStageApp: SSA)(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[(List[Instance], List[AMI])] = {
     for {
       prodInstances <- getInstances(stackStageApp)
       amiAttempts = amiArns(prodInstances).map(getAMI)
       amis <- Attempt.successfulAttempts(amiAttempts)
-    } yield instanceAmis(prodInstances, amis)
+    } yield (prodInstances, amis)
+  }
+
+  def instancesWithAmis(stackStageApp: SSA)(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[(Instance, Option[AMI])]] = {
```
adamnfish: It's just a question of taste so it's up to you, but this file is supposed to just contain the impure code that fetches data from Prism. The logic for handling the pure value that gets returned from Prism lives elsewhere. This helps to enforce the discipline of separating easily testable logic from code that depends on side effects and external systems.

`instancesWithAmis` and the other functions here now can't be tested without mocking the Prism API and providing an execution context. If they took a Prism response as an argument instead, you'd be able to test them by passing them a pure value and testing the pure value it returns. Of course, in these examples the code is so trivial that it's very likely you'll say they don't need tests. While this is fine and it's clear that there's no great harm caused by introducing these functions here, it comes back to the question of discipline. This starts the slippery slide of mixing logic with impure calls which will encourage future developers to add more features here, manipulating Prism responses in the same blocks that fetch those responses.

I'm happy to chat about this IRL, and definitely convince-able both ways, but thought it worth sharing the ideas behind the design as it stands.

app/prism/PrismLogic.scala
```diff
@@ -99,4 +101,26 @@ object PrismLogic {
       }
     }.getOrElse(true)
   }
+
+  /**
+    * @return Percentiles of instance AMIs age
```
adamnfish: Do these comments really add any value?

app/prism/PrismLogic.scala
```diff
@@ -99,4 +101,26 @@ object PrismLogic {
...
+    */
+  def instancesAmisAgePercentiles(amis: List[AMI]): Percentiles = {
+    val ages = amis.flatMap { ami =>
+      val amiAge = ami.creationDate.map(DateUtils.daysAgo).getOrElse(0)
```
adamnfish: I think a 'fairer' implementation would be to ignore ami's lacking a creation date, rather than assume they are brand-new. The former leaves the stats unchanged, this implementation moves us towards our goal without any evidence we've made progress. Better to flatten this, I think.

app/prism/PrismLogic.scala
```diff
@@ -99,4 +101,26 @@ object PrismLogic {
...
+      List.fill(ami.instances.getOrElse(Nil).length)(amiAge)
+    }
+    Percentiles(ages)
+  }
+
+  /**
+    * @return the number of instances from the list of AMIs that match a given SSA
+    */
+  def instancesCountForSSA(amis: List[AMI], ssa: SSA): Int = {
```
adamnfish: There's nothing in the signature that says these amis have to have been "hydrated" with their instances, which seems dangerous. If this still seems right then perhaps that's worth a comment.

Also, is this used?

app/utils/Stats.scala
```diff
@@ -0,0 +1,19 @@
+package utils
+
+case class Percentiles(values: Seq[Int]) {
+
+  private def percentile(s: Seq[Int], p: Double): Int = {
```
adamnfish: Strange choice to use a Double for a percentile! Why not Ints in the range [0, 100]?

app/views/fragments/metricsHeader.scala.html
```diff
@@ -0,0 +1,23 @@
+@(oldInstancesCount: Int, totalInstancesCount: Int, agePercentiles: utils.Percentiles)
+
+@percentileRow(percentileKey: Int, percentileValue: Int) = {
```
adamnfish: Should this go in the [fragments](https://github.com/guardian/amiable/tree/master/app/views/fragments) package?

test/prism/PrismLogicTest.scala
```diff
@@ -145,40 +149,48 @@ class PrismLogicTest extends FreeSpec with Matchers {
       val i2 = instanceWithSSA("i2", ssa2)
 
       "an SSA with an AMI" in {
-        amiSSAs(List(a1 -> List(i1))) shouldEqual Map(ssa1 -> List(a1))
+        val ami = a1.copy(instances = Some(List(i1)))
+        allSSAs(List(ami)) shouldEqual Map(ssa1 -> List(ami))
```
adamnfish: The containing block's text label no longer matches the name of the function you're testing. (Currently on line 141)

app/controllers/AMIable.scala
```diff
@@ -46,10 +55,7 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
         amis <- Attempt.successfulAttempts(amiArns.map(Prism.getAMI))
         amisWithUpgrades = amis.map(Recommendations.amiWithUpgrade(agents.allAmis))
         amisWithInstances = PrismLogic.amiInstances(amisWithUpgrades, instances)
-        amiSSAs = PrismLogic.amiSSAs(amisWithInstances)
-      } yield {
-        Ok(views.html.instanceAMIs(ssa, amisWithUpgrades, PrismLogic.sortSSAAmisByAge(amiSSAs)))
-      }
+      } yield Ok(views.html.instanceAMIs(ssa, amisWithInstances))
```
adamnfish: I've spotted that this is done in the template now. I think that if the template only needs them sorted one way, much better to do that here rather than have logic mixed up in the template as well. In an ideal world the template just converts its arguments into HTML.

I realise the template previous had some logic in it (it was a mix of here and there). I'm very sorry about that.

TBonnin: You are right. The list should be not be sorted in the template

app/prism/Prism.scala
TBonnin: I am not sure I understand what you mean here.
`instancesWithAmis` was already in this file, requiring an execution context and a config.
The logic that associate AMIs with instances or instances with AMIs (`instanceAmis` and `amiInstances`) still live in `PrismLogic`

app/prism/PrismLogic.scala
TBonnin: Makes sense. Will change

app/prism/PrismLogic.scala
TBonnin: Not used yet, but would be in a future PR.
You are right about the "hydrated" bit though. Any idea about how to improve the situation?

app/views/fragments/metricsHeader.scala.html
TBonnin: I see fragments as reusable pieces of template. Here a `percentileRow` has no meaning outside of this template which defined the rows container. It is only a code block used to avoid duplication.

adamnfish: Sorry, this already is in fragments, I'm being silly!

TBonnin:
@adamnfish 
I reverted a lot of my unecessary initial changes:
- Keeping Prism about fetching data (instances/amis) and keep logic and manipulations of this data into the controller
- Removing the instances field in AMI class. As you mention hydrating instances into AMIs led to issue because it was hard to tell if AMIs had been hydrated or not, and some function required it.

I also styled the representation of percentiles. Actually @zeftilldeath did all the work. Shoutout to me.

app/prism/PrismLogic.scala
```diff
@@ -99,4 +101,15 @@ object PrismLogic {
       }
     }.getOrElse(true)
   }
+
+  def instancesAmisAgePercentiles(amis: List[AMI]): Percentiles = {
```
TBonnin: 👍  Added tests d0b12da

PR #37, opened 2017-02-21
Store precentiles metrics in CloudWatch

TBonnin:
Not doing anything visually with those metrics yet

### DIFF
--- a/app/metrics/CloudWatch.scala
+++ b/app/metrics/CloudWatch.scala
@@ -49,13 +49,13 @@ object CloudWatch {
       .withValue(allStacks)
   )
 
-  private[metrics] def putRequest(metricName: String, count: Int): PutMetricDataRequest = {
+  private[metrics] def putRequest(metricName: String, value: Int): PutMetricDataRequest = {
     new PutMetricDataRequest()
       .withNamespace(namespace)
       .withMetricData {
         new MetricDatum()
           .withMetricName(metricName)
-          .withValue(count.toDouble)
+          .withValue(value.toDouble)
           .withDimensions(dimensions.asJava)
       }
   }
@@ -89,7 +89,7 @@ object CloudWatch {
   private def attemptWithRequest(request: GetMetricStatisticsRequest)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
     val tmp: Future[Either[AMIableErrors, Option[List[(DateTime, Double)]]]] = getWithRequest(request).map(ds => Right(Some(ds)))
     Attempt.fromFuture(tmp){ case e =>
-      Logger.warn("Failed to fetch old instance count cloudwatch data", e)
+      Logger.warn("Failed to fetch CloudWatch data", e)
       Right(None)
     }
   }
--- a/app/metrics/CloudWatch.scala
+++ b/app/metrics/CloudWatch.scala
@@ -86,17 +86,19 @@ object CloudWatch {
     awsToScala(client.putMetricDataAsync)(request)
   }
 
-  private def attemptWithRequest(request: GetMetricStatisticsRequest)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
-    val tmp: Future[Either[AMIableErrors, Option[List[(DateTime, Double)]]]] = getWithRequest(request).map(ds => Right(Some(ds)))
-    Attempt.fromFuture(tmp){ case e =>
+  def get(metricName: String)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
+    Attempt.fromFuture(getWithRequest(getRequest(metricName)).map(ds => Right(Option(ds)))){ case e =>
       Logger.warn("Failed to fetch CloudWatch data", e)
       Right(None)
     }
   }
 
-  def get(metricName: String)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
-    attemptWithRequest(getRequest(metricName))
+  def put(metricName: String, maybeValue: Option[Int]): Unit = {
+    maybeValue.fold {
+      Logger.warn(s"Not updating CloudWatch - no value available for '$metricName'")
+    }{ value =>
+      putWithRequest(putRequest(metricName, value))
+      Logger.debug(s"Updated CloudWatch metric '$metricName' with value '$value'")
+    }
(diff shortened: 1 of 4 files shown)


app/metrics/CloudWatch.scala
```diff
@@ -14,6 +14,16 @@ import scala.collection.JavaConverters._
...
 
 
+sealed abstract class CloudWatchMetric(val name: String)
+object CloudWatchMetrics {
+  case object OldCount extends CloudWatchMetric("instances-running-out-of-date-amis")
+  case object AmisAgePercentile25th extends CloudWatchMetric("instances-amis-percentile-25th")
+  case object AmisAgePercentile50th extends CloudWatchMetric("instances-amis-percentile-50th")
+  case object AmisAgePercentile75th extends CloudWatchMetric("instances-amis-percentile-75th")
+  case object AmisAgePercentile90th extends CloudWatchMetric("instances-amis-percentile-90th")
+  case object AmisAgePercentileHighest extends CloudWatchMetric("instances-amis-percentile-highest")
```
TBonnin: Not sure we need the `highest` metric. What do you think @adamnfish?

app/services/Agents.scala
```diff
@@ -79,21 +82,25 @@ class Agents @Inject()(amiableConfigProvider: AmiableConfigProvider, lifecycle:
     )
   }
 
-  def refreshOldProdInstanceCount(): Unit = {
+  def refreshInstancesInfo(): Unit = {
```
TBonnin: I could split this functions into 2 but that would mean querying the instances and AMIs twice at every refresh interval

app/metrics/CloudWatch.scala
```diff
@@ -51,33 +60,43 @@ object CloudWatch {
...
+  }
+
+  private def putWithRequest(request: PutMetricDataRequest) = {
+    awsToScala(client.putMetricDataAsync)(request)
   }
 
-  def attemptOldCountData(client: AmazonCloudWatchAsyncClient)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
-    val tmp: Future[Either[AMIableErrors, Option[List[(DateTime, Double)]]]] = getOldCountData(client).map(ds => Right(Some(ds)))
+  private def attemptWithRequest(request: GetMetricStatisticsRequest)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
+    val tmp: Future[Either[AMIableErrors, Option[List[(DateTime, Double)]]]] = getWithRequest(request).map(ds => Right(Some(ds)))
```
adamnfish: this type-annotated `tmp` is a mistake on my part. It's what I do while I'm still working out the implementation, I should have inlined it when I'd worked it out. Be good to clear that up here since this is changing anyway.

app/metrics/CloudWatch.scala
```diff
@@ -14,6 +14,16 @@ import scala.collection.JavaConverters._
...
 
 
+sealed abstract class CloudWatchMetric(val name: String)
+object CloudWatchMetrics {
+  case object OldCount extends CloudWatchMetric("instances-running-out-of-date-amis")
+  case object AmisAgePercentile25th extends CloudWatchMetric("instances-amis-age-percentile-25th")
+  case object AmisAgePercentile50th extends CloudWatchMetric("instances-amis-age-percentile-50th")
+  case object AmisAgePercentile75th extends CloudWatchMetric("instances-amis-age-percentile-75th")
+  case object AmisAgePercentile90th extends CloudWatchMetric("instances-amis-age-percentile-90th")
+  case object AmisAgePercentileHighest extends CloudWatchMetric("instances-amis-age-percentile-highest")
```
TBonnin: Not sure I need the `highest` metric. What do you think @adamnfish?

app/services/Metrics.scala
```diff
@@ -2,30 +2,36 @@ package services
...
 import rx.lang.scala.Observable
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
 
-
 @Singleton
 class Metrics @Inject()(environment: Environment, agents: Agents, lifecycle: ApplicationLifecycle) {
+
+  def saveMetric(metricName: String, maybeValue: Option[Int]): Unit = {
```
adamnfish: Would this be better placed in the CloudWatch object?

app/metrics/CloudWatch.scala
```diff
@@ -40,44 +49,54 @@ object CloudWatch {
...
 
-  def putOldCountRequest(count: Int): PutMetricDataRequest = {
+  private[metrics] def putRequest(metricName: String, value: Int): PutMetricDataRequest = {
     new PutMetricDataRequest()
       .withNamespace(namespace)
       .withMetricData {
         new MetricDatum()
           .withMetricName(metricName)
-          .withValue(count.toDouble)
+          .withValue(value.toDouble)
```
TBonnin: Any reason why we store a `Double` and not an `Int`?

app/services/Agents.scala
adamnfish: I think this looks reasonable.

app/metrics/CloudWatch.scala
```diff
@@ -40,44 +49,54 @@ object CloudWatch {
...
+  private def attemptWithRequest(request: GetMetricStatisticsRequest)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
+    val tmp: Future[Either[AMIableErrors, Option[List[(DateTime, Double)]]]] = getWithRequest(request).map(ds => Right(Some(ds)))
     Attempt.fromFuture(tmp){ case e =>
-      Logger.warn("Failed to fetch old instance count cloudwatch data", e)
+      Logger.warn("Failed to fetch CloudWatch data", e)
       Right(None)
     }
   }
+
+  def get(metricName: String)(implicit executionContext: ExecutionContext): Attempt[Option[List[(DateTime, Double)]]] = {
```
adamnfish: I think I'd consider inlining the `attemptWithRequest` function here, since it doesn't add any purity benefit having them be separate and it's only used once.

app/metrics/CloudWatch.scala
adamnfish: That represents the oldest instance AMI? No harm in keeping it since it's so simple?

app/metrics/CloudWatch.scala
adamnfish: Just that the SDK requires a double

PR #38, opened 2017-02-24
Show AMIs age percentile history

TBonnin:
- Charts on index page are shown in tabs.
- Each chart can now display multiple time series.
- Now showing 90 days of data rather than 60 (removing the data point dot to make the chart more readable

screen shot 2017-02-24 at 09 50 41
screen shot 2017-02-24 at 09 52 18

### DIFF
--- a/app/controllers/AMIable.scala
+++ b/app/controllers/AMIable.scala
@@ -4,6 +4,7 @@ import javax.inject.Inject
 
 import auth.AuthActions
 import config.AmiableConfigProvider
+import metrics.Charts
 import models._
 import play.api._
 import play.api.mvc._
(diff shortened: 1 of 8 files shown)


app/controllers/AMIable.scala
```diff
@@ -19,14 +20,23 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
 
   def index = AuthAction.async { implicit request =>
     val ssa = SSA(stage = Some("PROD"))
+    val charts = List(
```
adamnfish: I think this is too much logic (and hard-coded data) for the edge of the program. An example of why that's a good habit to try and stick to is that you can't currently test the construction of this list. I feel like a function that returns the charts when given the current values out the agents would be a nicer fit for here.

I could go either way on whether this actually needs any tests (possibly the map bit I guess), up to you.

public/stylesheets/main.css
```diff
@@ -380,4 +376,34 @@ h1 {
...
+    padding-top: 1em;
+}
+
+#charts-nav .tabs{
+    background-color: transparent;
+    font-size: 12px;
+}
+
+#charts-nav .tabs .tab {
+    text-transform: none;
```
adamnfish: What does this achieve?

public/stylesheets/main.css
```diff
@@ -380,4 +376,34 @@ h1 {
...
-}
\ No newline at end of file
+}
+
+/* Index charts */
+#charts-container {
+    padding-top: 1em;
+}
+
+#charts-nav .tabs{
```
TBonnin: You are right. I will stick to BEM-style classes

public/stylesheets/main.css
TBonnin: `.tabs` is actually a class provided by Materialize. http://materializecss.com/tabs.html
I was overriding the style to avoid uppercase for instance

app/utils/DateUtils.scala
```diff
@@ -6,6 +6,9 @@ import org.joda.time.{DateTime, Days}
 import play.api.libs.json.{Format, Reads, Writes}
 
 object DateUtils {
+
+  val freshnessLimit = 30
```
TBonnin: It is exposed and can be used elsewhere. In this case, it is used to show the green goal threshold in the percentiles graph. It is not absolutely necessary though

adamnfish: Yes! Spotted it up there but forgot to delete this comment, sorry :-)

TBonnin:
@adamnfish I think I addressed all your comments. Thank you again

PR #40, opened 2017-02-27
Show instances count for each AMIs

TBonnin:
screen shot 2017-02-24 at 17 54 00

Regarding the UI: @zef mentioned that it would be better if the icons (age and machine) is before the number.


app/prism/PrismLogic.scala
```diff
@@ -107,4 +107,18 @@ object PrismLogic {
...
   }
 
+  /**
+    * T will either be an AMI, or an AMI attempt.
+    * From a full list of Ts and instances and a list of SSAs, return unique
+    * SSA and AMI combinations with their respective number of instances
+    */
+  def instancesCountPerSSA[T](amisWithInstances: List[(T, List[Instance])], ssas: List[SSA]): Map[(SSA, T), Int] = {
+    for {
+      (t, instances) <- amisWithInstances.toMap
```
thomaska: Sorry but why is `toMap` required here?

test/models/InstanceTest.scala
```diff
@@ -0,0 +1,40 @@
+package models
+
+import org.scalatest.{FreeSpec, Matchers}
+import util.Fixtures._
+
+class InstanceTest extends FreeSpec with Matchers {
```
thomaska: 👍 🎆  Great job on writing tests!!!

app/prism/PrismLogic.scala
```diff
@@ -107,4 +107,18 @@ object PrismLogic {
     Percentiles(ages)
   }
 
+  /**
+    * T will either be an AMI, or an AMI attempt.
```
thomaska: To be honest seeing "either" here made me expect an `Either[T]` 😅

app/prism/PrismLogic.scala
TBonnin: It is because `amisWithInstances` is a list and he return type of the function is a `Map`. 
The type of the 1st generator in a `for` comprehension would be the type of the yielded value.

app/controllers/AMIable.scala
```diff
@@ -56,12 +56,15 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
           totalInstancesCount = instances.length,
           PrismLogic.instancesAmisAgePercentiles(instancesWithAmis)
         )
+        allSSAs = PrismLogic.instanceSSAs(instances) :+ ssa
```
adamnfish: Is there a reason not to prepend `ssa`?

```
ssa :: PrismLogic.instanceSSAs(instances)
```

app/models/models.scala
```diff
@@ -52,6 +52,9 @@ case class Instance(
   val amiArn = specification.get("imageArn")
 
   override def toString: String = s"Instance<$arn>"
+  def belongsToSSA(ssa: SSA): Boolean = ssa.stack == stack &&
```
adamnfish: I try very hard to avoid mixing logic with data structure definitions with my own code. The only time I allow myself exceptions is when there is a property that is inherently a part of the object itself, which is to say it naturally arises as a part of the instance itself. It follows that this is only possible if it does't take any parameters.
One rule of thumb is that you shouldn't need any tests for the models package, which makes sense if it's just modelling data.

This is my taste so you're very welcome to disagree, but it looks to me like this would fit nicely with the rest of the logic?

test/models/InstanceTest.scala
adamnfish: `Instance` of what? These are testing `Stack`s at the moment, I think `StackTest` would be a batter name. That said, as I mentioned above, tests on model instances is a sign that data definitions and logic have been muddled together.

test/prism/PrismLogicTest.scala
```diff
@@ -265,4 +265,32 @@ class PrismLogicTest extends FreeSpec with Matchers {
       percentiles.p90 shouldEqual Some(14)
     }
   }
+
+  "instancesCountPerSSA" - {
+    "returns " in {
```
adamnfish: This won't be very helpful in a test output log when it fails. How about "returns the expected result", or perhaps more helpfully, "correctly generates and associates the instance count with each SSA"?

test/models/InstanceTest.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  val i1 = emptyInstance("i1").copy(stack = Some("stack1"), stage = Some("stage1"), app = List("app1"))
+
+  "Instance" - {
+    "should belong to SSA with same stack name" in {
```
adamnfish: FYI, you can nest blocks using FreeSpec, which would make these messages less boiler-platey.

```scala
"Stack instance" - {
  "belongs to SSA" - {
    "with same stack name" in {
      val stack = SSA(stack = i1.stack)
      i1.belongsToSSA(stack) should be (true)
    }
    "with same stack and stage" in {
      val stack = SSA(stack = i1.stack, stage = i1.stage)
      i1.belongsToSSA(stack) should be (true)
    }
  }
  "does not belong to SSA" - {
    // etc
```

It's also extremely helpful for sharing test setup using block scope (not relevant in this case but worth keeping in mind). This is why I favour FreeSpec, it's incredibly flexible and powerful and doesn't depend on magic context like the "**it** should" pattern in many other matcher choices.

app/controllers/AMIable.scala
TBonnin: no reason. I will prepend it

app/models/models.scala
TBonnin: Your approach make sense. I will move `belongsToSSA` to PrismLogic

test/prism/PrismLogicTest.scala
TBonnin: Definitely an oversight. Thanks for noticing.

TBonnin:
@adamnfish Regarding the layout: I had to change a break point to only show 2 pills per row on screen < 1200px. https://github.com/guardian/amiable/pull/40/commits/51b7db9227cc3059ab00e5d7d3ff52049dbb3139

Large screen (>1200px): 3 pills per row
screen shot 2017-02-27 at 12 41 23

Medium screen (>600px and <1200px): 2 pills per row
screen shot 2017-02-27 at 12 41 37

Small screen (<600px): 1 pill per row
screen shot 2017-02-27 at 12 25 15

public/stylesheets/main.css
```diff
@@ -122,7 +122,7 @@ h1 {
     }
 }
 
-@media only screen and (min-width: 720px) {
+@media only screen and (min-width: 1200px) {
```
adamnfish: I missed this, very good.

test/prism/PrismLogicTest.scala
```diff
@@ -265,4 +265,69 @@ class PrismLogicTest extends FreeSpec with Matchers {
...
+        (ssa2, a1) -> 1,
+        (ssa3, a1) -> 1,
+        (ssa3, a2) -> 2
+      )
+      instancesCountPerSsaPerAmi(amisWithInstances, allSSAs) should be (expectedResult)
+    }
+  }
+
+
+  "Instance" - {
```
adamnfish: I think we can still do a bit better than this :-)

app/controllers/AMIable.scala
```diff
@@ -56,12 +56,15 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
           totalInstancesCount = instances.length,
           PrismLogic.instancesAmisAgePercentiles(instancesWithAmis)
         )
+        allSSAs = ssa :: PrismLogic.instanceSSAs(instances)
```
adamnfish: Why do we need to add the requested SSA to the list? `instances` came from searching for that AMI so either it's already there, or there aren't any instances with the ssa.

adamnfish:
Looks good to me!

test/prism/PrismLogicTest.scala
TBonnin: @adamnfish is this better? https://github.com/guardian/amiable/pull/40/commits/abc0fb6776b3da70041da9257ef01f00746f88b2 😉

app/controllers/AMIable.scala
TBonnin: @adamnfish Because we need the count for the requested SSA. (Top group of AMI pills).

The instance SSAs are usually 'full' SSAs with stack, stage and app values. On the other hand the requested SSA is usually a 'filtered' SSA with potentially `None` value for stage and app, which we consider like matching all instances/AMIs.
In ordered to get the instance count for the requested SSA I am adding it to the list.
An other solution could be to compute all possible values for SSAs, including SSAs with empty values (matching all instances/AMIs) but I found this solution would be adding a lot of complexity.

PR #42, opened 2017-03-15
Add Ami usage info

thomaska:
Add information about which instances are using a specific AMI.
This information is visible in the AMI details view:
screen shot 2017-03-15 at 18 05 36

In more detail:
screen shot 2017-03-15 at 18 05 54

I am not sure about if the layout/UI is the best :/

Many many thanks to @adamnfish for his help and functional Scala crash course!


app/prism/PrismLogic.scala
```diff
@@ -116,12 +116,12 @@ object PrismLogic {
     for {
       (t, instances) <- amisWithInstances.toMap
       ssa <- ssas
-      instancesCount = instances.filter(i => doesInstanceBelongToSSA(i, ssa)).length
-      if(instancesCount > 0)
+      instancesCount = instances.count(i => doesInstanceBelongToSSA(i, ssa))
```
thomaska: These were IntelliJ suggestions...I hope it's ok that I added them?

app/prism/Prism.scala
```diff
@@ -31,6 +31,18 @@ object Prism {
 
   def getInstances(stackStageApp: SSA)(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[Instance]] = {
     val url = instancesUrl(stackStageApp, config.prismUrl)
+    getInstancesFromUrl(url)
+  }
+
+  def imageUsage(imageOpt: Option[AMI])(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[Instance]] = {
+    for {
+      image <- Attempt.fromOption(imageOpt, AMIableErrors(AMIableError("Could not find image", "Could not find image", 404)))
```
thomaska: I am not sure if those messages are the best?...and also since image is an `Option`, would it make sense trying to print its id?

TBonnin: Yep adding the image id in the error message could be useful

app/prism/PrismLogic.scala
TBonnin: that makes sense

app/prism/PrismLogic.scala
```diff
@@ -116,12 +116,12 @@ object PrismLogic {
...
+      if instancesCount > 0
```
TBonnin: I am not sure I like this syntax to be honest. I don't feel strongly though so up to you

app/views/fragments/instanceTable.scala.html
```diff
@@ -0,0 +1,50 @@
...
+                <td class="instance-details-table__entry">@instance.name</td>
+                <td class="instance-details-table__entry">@instance.arn</td>
+                <td class="instance-details-table__entry">@instance.dnsName</td>
+                <td nowrap class="instance-details-table__entry">@instance.group</td>
+                <td class="instance-details-table__entry">@instance.ip</td>
+                <td class="instance-details-table__entry">@instance.createdAt</td>
+                <td class="instance-details-table__entry">@instance.instanceName</td>
+                <td nowrap class="instance-details-table__entry">@instance.region</td>
+                <td class="instance-details-table__entry">@instance.vendor</td>
+                <td class="instance-details-table__entry">@instance.specification.getOrElse("instanceType", emptyChar)</td>
```
TBonnin: I don't think we need all those details. I would vote to keep `stack, stage, app, instanceName, createdAt` and maybe `dnsName or IP`. Everything else seems superfluous. What do you think?

app/prism/Prism.scala
```diff
@@ -31,6 +31,18 @@ object Prism {
 
   def getInstances(stackStageApp: SSA)(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[Instance]] = {
     val url = instancesUrl(stackStageApp, config.prismUrl)
+    getInstancesFromUrl(url)
+  }
+
+  def imageUsage(imageOpt: Option[AMI])(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[Instance]] = {
```
TBonnin: Why is the argument an Optional? What about dealing with getting or not an AMI outside of this function?

app/controllers/AMIable.scala
```diff
@@ -42,9 +42,10 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
     attempt {
       for {
         amis <- Prism.getAMIs()
-        ami = amis.find(_.imageId == imageId)
-        amiWithUpgrade = ami.map(Recommendations.amiWithUpgrade(agents.allAmis))
-      } yield Ok(views.html.ami(amiWithUpgrade))
+        maybeAmi = amis.find(_.imageId == imageId)
+        instances <- Prism.imageUsage(maybeAmi)
+        amiWithUpgrade = maybeAmi.map(Recommendations.amiWithUpgrade(agents.allAmis))
+      } yield Ok(views.html.ami(amiWithUpgrade, instances))
```
TBonnin: We also need to take `launch configurations` into account. An AMI could be for example used in a `launch configuration` with an ASG that has currently 0 instances.
This might be added in a following pull request though

app/views/fragments/instanceTable.scala.html
```diff
@@ -0,0 +1,50 @@
+@(instances: List[Instance], emptyChar: String = "-")
+
+<div class="container">
+@if(instances.nonEmpty) {
+    <div class="ami-usage--message">AMI Usage</div>
```
TBonnin: I would add here the instances and launch configuration count. 
Ex: `This AMI is used by 8 instances and 3 launch configurations`. Something like this

public/stylesheets/main.css
```diff
@@ -243,6 +243,25 @@ h1 {
...
 
+.ami-usage--message {
+    font-size: x-large;
+    color: #3661B0;
+    margin-top: 30px;
+}
+
+.instance-details__container {
+    display: flex;
+}
```
TBonnin: I don't see this class being used. Am I wrong?

thomaska:
[Get a list of instances running that AMI (group by stack/account)](https://trello.com/c/tnFFyHeO/21-get-a-list-of-instances-running-that-ami-group-by-stack-account)

app/controllers/AMIable.scala
thomaska: Yes you are right, I will create a separate PR for this!

thomaska:
Thanks a lot for the review @TBonnin !! Really good spots!

The updated view looks like this:
screen shot 2017-03-16 at 13 02 07

app/controllers/AMIable.scala
```diff
@@ -42,12 +42,19 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
     attempt {
       for {
         amis <- Prism.getAMIs()
-        ami = amis.find(_.imageId == imageId)
-        amiWithUpgrade = ami.map(Recommendations.amiWithUpgrade(agents.allAmis))
-      } yield Ok(views.html.ami(amiWithUpgrade))
+        ami <- extractAmi(imageId, amis)
+        amiWithUpgrade = Recommendations.amiWithUpgrade(agents.allAmis)(ami)
+        instances <- Prism.imageUsage(ami)
+        launchConfig <- Prism.launchConfigUsage(ami)
```
TBonnin: I would pluralize `laucnConfig` here

TBonnin:
Is the result ordered/grouped by `stack`?

app/views/fragments/amiUsage.scala.html
```diff
@@ -0,0 +1,62 @@
...
+        <tbody>
+        @for(instance <- instances) {
+            <tr>
+                <td class="instance-details-table__entry">@instance.stack.getOrElse(emptyChar)</td>
+                <td class="instance-details-table__entry">@instance.stage.getOrElse(emptyChar)</td>
+                <td class="instance-details-table__entry">@instance.app</td>
+                <td class="instance-details-table__entry">@instance.vendorState</td>
+                <td class="instance-details-table__entry">@instance.instanceName</td>
+                <td class="instance-details-table__entry">@instance.ip</td>
+                <td class="instance-details-table__entry">@instance.createdAt</td>
```
sihil: Can this use a date formatter to make the UI slightly nicer?

app/views/fragments/amiUsage.scala.html
```diff
@@ -0,0 +1,62 @@
...
+    <div class="ami-usage--message">Instances</div>
+    <table class="striped ami-usage--table">
+        <thead>
+            <tr>
+                <th class="instance-details-table__heading">Stack</th>
+                <th class="instance-details-table__heading">Stage</th>
+                <th class="instance-details-table__heading">App</th>
+                <th class="instance-details-table__heading">State</th>
+                <th class="instance-details-table__heading">Instance name</th>
+                <th class="instance-details-table__heading">Ip</th>
```
sihil: Nit-pick, I'd write this as IP as it is an acronym.

app/views/fragments/amiUsage.scala.html
```diff
@@ -0,0 +1,62 @@
...
+            <tr>
+                <td class="instance-details-table__entry">@launchConfig.meta.origin.accountName</td>
+                <td class="instance-details-table__entry">@launchConfig.name</td>
+                <td class="instance-details-table__entry">@launchConfig.createdTime</td>
+                <td class="instance-details-table__entry">@launchConfig.instanceType</td>
+            </tr>
+        }
+        </tbody>
+    </table>
+}
```
sihil: This is a thought rather than a request: Could launch configs go above instances? It should always be a shorter list so won't become buried at the bottom if first. Would that make more sense??

app/views/fragments/amiUsage.scala.html
thomaska: Is this format ok? `dd-MM-yyyy HH:mm:ss`

sihil: Sure, you could also drop the seconds as far as I'm concerned.

thomaska: Eventually I reused an existing format, hope its ok!

thomaska:
Updated (2) view:
screen shot 2017-03-20 at 17 56 23

thomaska:
@TBonnin I added also the sorting as you suggested!

app/prism/PrismLogic.scala
```diff
@@ -124,4 +124,25 @@ object PrismLogic {
...
+
+  /**
+    * Sorts Instances by Stack, App, Stage in ascending order
+    */
+  def sortInstancesByStack: List[Instance] => List[Instance] = {
+    insts => insts.sortWith {
+      (i1, i2) => i1.stack.getOrElse("") < i2.stack.getOrElse("") ||
+        i1.app.headOption.getOrElse("") < i2.app.headOption.getOrElse("") || i1.stage.getOrElse("") < i2.stage.getOrElse("")
+    }
+  }
```
TBonnin: Would it make sense to define an `Ordering` in the companion object of `Instance` and `LaunchConfiguration` and then use the `.sorted` method?

app/controllers/AMIable.scala
```diff
@@ -46,7 +46,10 @@ class AMIable @Inject()(override val amiableConfigProvider: AmiableConfigProvide
         amiWithUpgrade = Recommendations.amiWithUpgrade(agents.allAmis)(ami)
         instances <- Prism.imageUsage(ami)
         launchConfigs <- Prism.launchConfigUsage(ami)
-      } yield Ok(views.html.ami(amiWithUpgrade, instances, launchConfigs))
+      } yield Ok(views.html.ami(
+        amiWithUpgrade,
+        PrismLogic.sortInstancesByStack(instances),
+        PrismLogic.sortLCsByOwner(launchConfigs)))
```
TBonnin: Would it make sense to define an `Ordering` in the companion object of `Instance` and `LaunchConfiguration` and then use `.sorted`

thomaska: Yes that is a very good option! I did it like that because I felt it was consisted with `Prismlogic.sortSSAAmisByAge` but I am happy to refactor!

test/prism/PrismLogicTest.scala
```diff
@@ -330,4 +330,51 @@ class PrismLogicTest extends FreeSpec with Matchers {
       }
     }
   }
-}
\ No newline at end of file
+
+  "sortInstancesByStack" - {
+    "should sort by stack, app, stage" - {
+      "when stack and stage are non empty" in {
+        val in1 = Instance("arn", "name", "state", "group", "dns","ip", DateTime.now(), "instanceName", "region", "vendor", List.empty,Map.empty, Some("stack1"), Some("stageB"), List("appB"), List.empty, Map.empty, null)
```
adamnfish: It's good to try and keep the actual test cases as succinct as possible. This is fine as is, but when I find myself writing the same test boiler-plate over and over again I normally try and pull it out into an enclosing block as shared setup. Then you only need to tweak it as needed in each test case.

e.g. (disclaimer: highly contrived example)

```scala
"function I'm testing" - {
  val thing = MyThing("lengthy", "setup", "here", 1, 2, 3, etc)

  "when stack is empty" - {
    val thingWithEmptyStack = thing.copy(stack = "")

    "returns false when this condition is true" in {
      myFunction(thingWithEmptyStack, other, args) shouldEqual false
    }

    "returns true in this case" in {
      myFunction(thingWithEmptyStack, different, args) shouldEqual true
    }
  }

  "when stack has some other property" - {
    etc
  ...
}
```

This is why I prefer to use `FreeSpec`. It lets you arbitrarily create blocks that have helpful labels on them and then at the bottom you have the actual test case, which is marked with `in`. The message that the test runner gets is the full path to the actual test case e.g. `"function I'm testing when stack is empty returns false when this condition is true"` (contrived example, but hopefully you see what I mean). you can group your tests with shared setup and the test messages when they run make sense.

With this in mind if you have a look around the existing tests you'll see why they're laid out the way they are. It's about grouping tests in a way that makes them easy to understand for the reader because the use case of comprehending existing test cases is the most important one to address when writing tests.

test/prism/PrismLogicTest.scala
```diff
@@ -330,4 +330,51 @@ class PrismLogicTest extends FreeSpec with Matchers {
       }
     }
   }
-}
\ No newline at end of file
+
+  "sortInstancesByStack" - {
+    "should sort by stack, app, stage" - {
```
adamnfish: This nesting doesn't do anything. As-is I'd combine these into a single block, but see what you think after you've read the more lengthy description I've put below.

test/prism/PrismLogicTest.scala
```diff
@@ -330,4 +330,51 @@ class PrismLogicTest extends FreeSpec with Matchers {
...
+        val in1 = Instance("arn", "name", "state", "group", "dns","ip", DateTime.now(), "instanceName", "region", "vendor", List.empty,Map.empty, None, None, List("app1"), List.empty, Map.empty, null)
+        val in2 = in1.copy(app = List("app2"))
+        val instanceList = List(in2, in1)
+        PrismLogic.sortInstancesByStack(instanceList) should contain inOrderOnly(in1, in2)
+      }
+    }
+  }
+
+  "sortLCsByOwner" - {
+    "should sort Launch Configurations" - {
```
adamnfish: Again, this nesting doesn't achieve much. I'll generally only introduce a block where there's a reason for doing so, for example to introduce some block-scoped test data or as a helpful description of a group of tests.

test/util/Fixtures.scala
```diff
@@ -119,4 +119,34 @@ object Fixtures {
                           |  }
                           |}""".stripMargin
   }
+
+  object LaunchConfigs {
+    val validLaunchConfiguration =
+      """{
+        |        "name": "BackpageCODE-LaunchConfig-1GEMXJYENZ5VECopy",
```
adamnfish: Is this fake test data?

test/prism/JsonUtilsTest.scala
```diff
@@ -100,4 +100,28 @@ class JsonUtilsTest extends FreeSpec with Matchers with EitherValues with Attemp
       extractInstance(json).awaitEither.isLeft shouldBe true
     }
   }
+
+  "extractLaunchConfiguration" - {
+    "should return a Launch Configuration from correct json" in {
+      val json = Json.parse(LaunchConfigs.validLaunchConfiguration)
+      val launchConfig = extractLaunchConfiguration(json).awaitEither.right.value
+      launchConfig should have(
+        'name ("BackpageCODE-LaunchConfig-1GEMXJYENZ5VECopy"),
```
thomaska: 😿

thomaska:
@adamnfish maybe could have another look/review?!
It would be nice if we'd manage to merge this before our tomorrow's presentation/demo with @TBonnin !

adamnfish:
Looks great!

thomaska:
thanks! 😄

PR #43, opened 2017-04-11
Team notifications

thomaska:
Notify the owner of each SSA (Stack-Stage-App combination) about its status.

This runs a cron job that sends an email every Friday afternoon (parametrized) to the Owners that have instances running with old AMIs (and only them!).

The email looks like this:
screen shot 2017-05-03 at 18 08 27

It must be noted that if there is an owner of a Stack-Stage-App combination and an owner of the whole Stack - both of them will receive an email.
(eg the owner of `security-janus-PROD` and the owner of `security` will both receive emails about janus if it runs on old instances!)

~~This PR uses a Prism endpoint that is in this PR: 
https://github.com/guardian/prism/pull/44
(so it is dependent on that being merged!)~~


app/metrics/CloudWatch.scala
```diff
@@ -27,13 +27,13 @@ object CloudWatchMetrics {
...
     val credentialsProvider = new AWSCredentialsProviderChain(
-      new InstanceProfileCredentialsProvider(),
+      InstanceProfileCredentialsProvider.getInstance(),
       new ProfileCredentialsProvider("deployTools")
     )
     val region = Option(Regions.getCurrentRegion).getOrElse(Region.getRegion(Regions.EU_WEST_1))
-    val acwac = new AmazonCloudWatchAsyncClient(credentialsProvider)
-    acwac.setRegion(region)
-    acwac
+    val acwac = AmazonCloudWatchAsyncClientBuilder.standard().withCredentials(credentialsProvider)
```
thomaska: this is because of the update in aws library used!

app/models/models.scala
```diff
@@ -130,3 +132,18 @@ object LaunchConfiguration {
   import utils.DateUtils._
   implicit val jsonFormat = Json.format[LaunchConfiguration]
 }
+
+case class Email(address: String, subject: String, message: String)
```
thomaska: yes there is `SendEmailRequest`
It has the following ctor:
```
public SendEmailRequest(String source, Destination destination, Message message)
```
We could move the stuff that are being done in the `AWSMailClient.send` function and actually move around a `SendEmailRequest class. The thought behind creating the `Email` class was that we might change (?) mail client library in the future so it would be better to have generic (and custom made) classes being moved around...but maybe this is not very probable!

app/services/notification/NotificationScheduler.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import org.quartz.CronScheduleBuilder.cronSchedule
+import org.quartz.JobBuilder.newJob
+import org.quartz.TriggerBuilder.newTrigger
+import org.quartz.{JobKey, TriggerKey}
+import org.quartz.impl.StdSchedulerFactory
+import play.api.Logger
+
+object NotificationScheduler {
+  private val notifyOwnersSchedule = "0 0 16 ? * FRI *"
```
thomaska: Yes good shout! It should!

app/services/notification/NotificationScheduler.scala
```diff
@@ -0,0 +1,33 @@
+package services.notification
+
+import org.quartz.CronScheduleBuilder.cronSchedule
+import org.quartz.JobBuilder.newJob
+import org.quartz.TriggerBuilder.newTrigger
+import org.quartz.{JobKey, TriggerKey}
+import org.quartz.impl.StdSchedulerFactory
+import play.api.Logger
+
+object NotificationScheduler {
```
thomaska: Yes you are right! Will amend!

app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -0,0 +1,37 @@
+package services.notification
+
+import config.AMIableConfig
+import models.{Email, SSA}
+import play.api.Logger
+import prism.{Prism, PrismLogic}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ScheduledNotificationRunner {
```
thomaska: :-) Yes this is indeed WIP!

app/services/notification/mailClient.scala
```diff
@@ -0,0 +1,48 @@
...
+import com.amazonaws.services.simpleemail.model._
+import com.google.inject.ImplementedBy
+import models.Email
+import play.api.{Configuration, Logger}
+import play.api.libs.concurrent.Execution.Implicits._
+
+import scala.concurrent.{Future, Promise}
+
+@ImplementedBy(classOf[AWSMailClient])
+trait MailClient {
```
thomaska: The idea (as mentioned also in a previous comment) is to abstract out the implementation library for sending the emails in case we would like to change it? 
Maybe this is causing noise though rather than helping, I don't know. :/

app/prism/Prism.scala
```diff
@@ -11,6 +11,15 @@ import scala.concurrent.ExecutionContext
 object Prism {
   import PrismLogic._
 
+  def getOwners(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[Owner]] = {
+    val url = ownersUrl(config.prismUrl)
+    for {
+      response <- Http.response(config.wsClient.url(url).get(), "Unable to fetch Owners", url)
+      jsons <- ownersResponseJson(response)
+      owners <- Attempt.sequence(jsons.map(extractOwner))
```
thomaska: I will amend as soon as this gets merged!

thomaska:
@adamnfish @sihil Could you maybe have a look whenever you would have some time!? 😅

app/services/notification/NotificationScheduler.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.quartz.CronScheduleBuilder.cronSchedule
+import org.quartz.JobBuilder.newJob
+import org.quartz.TriggerBuilder.newTrigger
+import org.quartz.{JobKey, TriggerKey}
+import org.quartz.impl.StdSchedulerFactory
+import play.api.Logger
+
+import scala.concurrent.ExecutionContext
+
+class NotificationScheduler {
```
adamnfish: Is this *meaningfully* different to the `Notifications` class - what does the Notifications class (/service) *do* if not schedule notifications? Is there a reason why they need to be separate?

It might be simpler as one class. Also, combining them into a single Guice-y class would go some way towards solving the multiple instance thing I mention below.

Looking for a conversation here (as ever!), I'm certainly not just saying "do what I think" :-)

app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -0,0 +1,40 @@
...
+import config.AMIableConfig
+import models.{Instance, Owner, SSA}
+import play.api.Logger
+import prism.{Prism, PrismLogic}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ScheduledNotificationRunner {
+
+  def run(mailClient: AWSMailClient)(implicit config: AMIableConfig, ec: ExecutionContext): Future[Any] = {
```
adamnfish: `Any`? :-(

I think `Unit` would be better here. `Any` doesn't tell the reader (or compiler) anything. `Unit` documents that you're running this method for its side-effects, in this case "send an email" or "log an error".

adamnfish: That said, if you use `Attempt` throughout the program instead of mixing it with Futures you can write this really neatly with a for-comp. I had a fiddle and came up with this:

```scala
  def run(mailClient: AWSMailClient)(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[String]] = {
    for {
      instancesWithAmis <- Prism.instancesWithAmis(SSA(stage = Some("PROD")))
      oldInstances = PrismLogic.oldInstances(instancesWithAmis)
      _ = Logger.debug(s"Found ${oldInstances.size} PROD instances running on an out-of-date AMI")
      owners <- Prism.getOwners
      mailIds <- Attempt.traverse(owners)(owner => mailClient.send(owner, instancesForOwner(owner, oldInstances)))
    } yield mailIds
  }
```

This required changing `AWSMailClient.send` to return an `Attempt[String]` instead of a `Future[String]` like it does now.

Most of the error logs duplicate logs that are written elsewhere in the program so they probably aren't needed. I've left the debug one in the example!

app/services/notification/AWSMailClient.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.amazonaws.services.simpleemail.model._
+import models.{Instance, Owner}
+import play.api.libs.concurrent.Execution.Implicits._
+import play.api.{Configuration, Logger}
+
+import scala.concurrent.{Future, Promise}
+
+class AWSMailClient @Inject()(amazonMailClient: AmazonSimpleEmailServiceAsync, configuration: Configuration) {
+
+  def send(owner: Owner, instances: Seq[Instance]): Future[String] = {
```
adamnfish: Am I right to think this will send even if there are no old instances to mention? Is this the expected behaviour? This should probably just do as its told, I think the logic for that should probably be elsewhere but I spotted it here so I'll mention it here.

app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -0,0 +1,40 @@
...
+      { instancesWithAmis =>
+        val oldInstances = PrismLogic.oldInstances(instancesWithAmis)
+        Logger.debug(s"Found ${oldInstances.size} PROD instances running on an out-of-date AMI")
+        Prism.getOwners.fold ({
+          err =>
+            Logger.warn(s"Failed to get Owners ${err.logString}")
+        },{
+          owners =>
+            owners.map { owner =>
+              mailClient.send(owner, instancesForOwner(owner, oldInstances))
```
adamnfish: As I mentioned in the `send` bit, this looks like it sends even if there are no old instances to alert about. Is that what we want?

build.sbt
```diff
@@ -35,8 +35,10 @@ libraryDependencies ++= Seq(
   ws,
   "com.typesafe.akka" %% "akka-agent" % "2.4.2",
   "io.reactivex" %% "rxscala" % "0.26.0",
-  "com.amazonaws" % "aws-java-sdk-cloudwatch" % "1.10.67",
+  "com.amazonaws" % "aws-java-sdk-cloudwatch" % "1.11.113",
```
adamnfish: :+2:

;-)

conf/application.conf
```diff
@@ -57,3 +57,5 @@ auth {
     2faGroupId=${2FA_GROUP_ID}
   }
 }
+
+amiable.mailClient.fromAddress = ${MAIL_ADDRESS}
```
adamnfish: Make sure we don't merge/deploy this PR without sorting the config out first.

app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -0,0 +1,40 @@
...
+            }
+        })
+      }
+    )
+  }
+
+  def instancesForOwner(owner: Owner, oldInstances: List[Instance]): List[Instance] = {
+    oldInstances.filter(i => {
+      owner.stacks.exists(ssa => ssa == SSA(i.stack, i.stage, i.app.headOption) || ssa == SSA(i.stack, stage = None, i.app.headOption) ||
```
adamnfish: Is it ok to take `app.headOption`, or will that mean we miss some stuff?

adamnfish: Just out of interest, how different would this behave in practice if the logic was just the following (I'm not sure how the Prism owners stuff works)

```scala
oldInstances.filter { i =>
  owner.stacks.exists(ssa => ssa.stage == i.stage)
}
```

app/services/notification/AWSMailClient.scala
```diff
@@ -0,0 +1,46 @@
...
+    val request: SendEmailRequest = createEmailRequest(owner, instances)
+
+    val promise = Promise[SendEmailResult]()
+    val responseHandler = new AsyncHandler[SendEmailRequest, SendEmailResult] {
```
sihil: +1

sihil:
In response to @adamnfish's thoughts about multiple notifications...

There are a couple of solutions to this. One might be to pull this code out into a lambda that is run on a schedule. Another might be to lookup which instances of amiable are currently running and only run it on the oldest / newest instance. The latter has some edge cases where you might wind up with a race condition if there is a deploy at the time the e-mail is scheduled, but is better than nothing.

app/services/notification/AWSMailClient.scala
thomaska: Yes you are right!...I thought this was already addressed, but apparently it was neglected!
I will amend!

app/services/notification/AWSMailClient.scala
```diff
@@ -0,0 +1,46 @@
...
+      override def onError(e: Exception): Unit = {
+        Logger.warn(s"Could not send mail: ${e.getMessage}\n" +
```
thomaska: Reusing the `awsToScala` method will remove this whole block, so there are gonna be the "default" messages:
`Logger.warn("Failed to execute AWS SDK operation", e)`

adamnfish: True, but you'll get the stacktrace there as well so it should be clear what went wrong

app/services/notification/AmazonSimpleEmailServiceAsyncFactory.scala
```diff
@@ -0,0 +1,18 @@
+package services.notification
+
+import javax.inject.Inject
+
+import com.amazonaws.regions.Regions
+import com.amazonaws.services.simpleemail._
+import com.google.inject.Provider
+
+
+class AmazonSimpleEmailServiceAsyncFactory @Inject()(awsCredentials :AwsCredentials) extends Provider[AmazonSimpleEmailServiceAsync]{
```
thomaska: Yes exactly...the gods required this! :godmode:  Maybe since there is the intention of moving to compile time DI maybe it is ok to leave it like that?!

app/services/notification/AWSMailClient.scala
adamnfish: (and if you're really worried you have a Future, so you can do what you want with that, including log when it fails. This is particularly true if you change this to Attempt since you'll need to handle the failure case then anyway)

app/services/notification/AwsCredentials.scala
```diff
@@ -0,0 +1,14 @@
+package services.notification
+
+import javax.inject.Singleton
+
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.auth.{AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain}
+
+@Singleton
+class AwsCredentials {
```
thomaska: A holder object for the Aws credential provider...it's being used in creating the `AmazonSimpleEmailServiceAsync`. Maybe this is not a very good way to do this?!
Thought to keep it in a separate place (rather that embedding it in the code creating the `AmazonSimpleEmailServiceAsync`)  so that it can be re-used...?

app/models/models.scala
```diff
@@ -130,3 +132,16 @@ object LaunchConfiguration {
...
+object Owner {
+  implicit val jsonFormat = Json.format[Owner]
+}
+
+/**
+  * @param quartzCronExpression a Quartz cron expression,
+  *                             e.g. "0 0 3 * * ? *" to run at 3am every day
+  *                             See http://www.quartz-scheduler.org/documentation/quartz-2.2.x/tutorials/crontrigger.html
+  */
+case class NotifySchedule(quartzCronExpression: String)
```
thomaska: yeah maybe, since this is only used in `NotificationScheduler` and not passed around, makes sense to remove it and only have a string. I will amend.

app/services/notification/AwsCredentials.scala
adamnfish: Provider is (essentially) static here so you can inline this without affecting anything. You'll see in (e.g.) `metrics.CloudWatch` how it's done elsewhere. Magic Dependency Injection makes it super easy to create 1000 classes where 1 will do, it takes some discipline to find the right balance between flexibility and simplicity. I prefer simplicity, especially in "microapps". Other opinions exist and are valid :-)

app/services/notification/ScheduledNotificationRunner.scala
adamnfish: Oops, I meant `stack` not `stage`.

app/services/notification/AWSMailClient.scala
adamnfish: Oh I misread this the first time and thought it was putting the email in the error! Yeah, what Simon said :-)

Sorry!

thomaska:
@adamnfish I committed the changes we did while pairing and tested it locally...it seems to be working!

app/models/Attempt.scala
```diff
@@ -81,6 +81,10 @@ object Attempt {
     Attempt(future recover recovery)
   }
 
+  def fromFuture2[A](future: Future[A])(recovery: PartialFunction[Throwable, Either[AMIableErrors, A]])(implicit ec: ExecutionContext): Attempt[A] = {
```
adamnfish: `fromFuture2` probably isn't the right name! `fromFuture` is fine - they have different signatures. We discussed in person, the other `fromFuture` looks a bit weird to me. I think this one is much better so I'll think about migrating the existing usages to this at some point.

app/prism/JsonUtils.scala
```diff
@@ -69,6 +69,12 @@ object JsonUtils {
     }
   }
 
+  def ownersResponseJson(response: WSResponse): Attempt[List[JsValue]] = {
+    JsonUtils.jsResultToAttempt("Could not get Launch Configuration from response JSON"){
```
adamnfish: I think this error message is erroneously copied from eslewhere?

app/services/notification/AmazonSimpleEmailServiceAsyncFactory.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.amazonaws.regions.{Region, Regions}
+import com.amazonaws.services.simpleemail._
+import com.google.inject.Provider
+
+
+class AmazonSimpleEmailServiceAsyncFactory extends Provider[AmazonSimpleEmailServiceAsync]{
+
+  private lazy val provider = new AWSCredentialsProviderChain(
+    new ProfileCredentialsProvider("deployTools"),
+    new DefaultAWSCredentialsProviderChain())
```
adamnfish: Maybe better to be explicit here? The `DefaultAWSCredentialsProviderChain` matches loads of stuff, including the local default profile. I've seen people get pretty confused and waste lots of time debugging Janus creds when an app is loading credentials they didn't expect. In this case we want the `InstanceProfileCredentialsProvider` in AWS and to fall back to the Janus profile for local development. We really don't want it to look anywhere else.

```scala
new AWSCredentialsProviderChain(
  new InstanceProfileCredentialsProvider(),
  new ProfileCredentialsProvider("deployTools")
)
```

app/services/notification/NotificationJob.scala
```diff
@@ -0,0 +1,33 @@
...
+    val schedulerContext = context.getScheduler.getContext
+    val mailClient = schedulerContext.get("MailClient").asInstanceOf[AWSMailClient]
+    implicit val amiableConfig = schedulerContext.get("AMIableConfig").asInstanceOf[AMIableConfig]
+    implicit val ec = schedulerContext.get("ExecutionContext").asInstanceOf[ExecutionContext]
+    implicit val configuration = schedulerContext.get("Configuration").asInstanceOf[Configuration]
+    val result = ScheduledNotificationRunner.run(mailClient)
+    Await.result(result.asFuture, 1.minute).fold(
+      err => Logger.error(s"Failed to send scheduled notifications: ${err.logString}"), { msgs =>
+        val (notSent, sent) = msgs.partition(_ == ScheduledNotificationRunner.MessageNotSent)
+        Logger.info(s"Sent ${sent.size} ${message_s(sent.size)}. ${notSent.size} ${message_s(notSent.size)} not sent")
```
adamnfish: `message_s` is cute, but is it necessary? Also, in a log file it might not be super obvious what this refers to. The following is an example of something a bit simpler that might be easier to understand in the log files?
```
Email notifications: ${notSent.size} failed, ${sent.size} sent
```

*minor point :-)

app/services/notification/Notifications.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.quartz.{JobKey, TriggerKey}
+import org.quartz.TriggerBuilder.newTrigger
+import org.quartz.impl.StdSchedulerFactory
+import play.api.inject.ApplicationLifecycle
+import play.api.{Configuration, Environment, Logger, Mode}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class Notifications @Inject()(
```
adamnfish: Bit out of whack layout-wise.

app/services/notification/Notifications.scala
```diff
@@ -0,0 +1,63 @@
...
+                               amiableConfigProvider: AmiableConfigProvider,
+                               environment: Environment,
+                               lifecycle: ApplicationLifecycle,
+                               mailClient: AWSMailClient)
+                             (implicit exec: ExecutionContext,
+                              configuration: Configuration) {
```
adamnfish: Why's there an `amiableConfigProvider` and a `configuration`?

app/services/notification/Notifications.scala
```diff
@@ -0,0 +1,63 @@
...
+    * a Quartz cron expression,
+    * e.g. "0 0 3 * * ? *" to run at 3am every day
+    * See http://www.quartz-scheduler.org/documentation/quartz-2.2.x/tutorials/crontrigger.html
+    */
+  private val ownerSchdlCron = configuration.getString("amiable.owner.notification.cron").get
+  private val scheduler = StdSchedulerFactory.getDefaultScheduler
+
+  // send notifications only in PROD
+  if (environment.mode == Mode.Prod) {
+    Logger.info("Starting the scheduler")
```
adamnfish: Might be worth saying what this scheduler is for in the log message.

Really minor stuff again!

app/services/notification/Notifications.scala
```diff
@@ -0,0 +1,63 @@
...
+      .build()
+    val trigger = newTrigger()
+      .withIdentity(triggerKey("notificationTrigger"))
+      .withSchedule(cronSchedule(ownerSchdlCron))
+      .build()
+    scheduler.scheduleJob(jobDetail, trigger)
+    Logger.info(s"Scheduled owner notification with schedule [$ownerSchdlCron]")
+  }
+
+  private def triggerKey(id: String): TriggerKey = new TriggerKey(id)
```
adamnfish: Do you feel like these two functions are offering any value?

app/views/email.scala.html
```diff
@@ -0,0 +1,71 @@
+@import utils.DateUtils
+@import views.html.fragments.heading
+
+@(instances: Seq[Instance], stack: Owner,emptyChar: String = "-")
```
adamnfish: Would `stack` be better named as `owner`?

app/services/notification/AWSMailClient.scala
```diff
@@ -0,0 +1,21 @@
+package services.notification
+
+import javax.inject.Inject
+
+import aws.AwsAsyncHandler.awsToScala
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceAsync
+import com.amazonaws.services.simpleemail.model._
+import models.{Attempt, _}
```
adamnfish: This is weird!

*super minor :-)

app/prism/JsonUtils.scala
thomaska: Yes! Bad copy-pasting!

app/services/notification/NotificationJob.scala
thomaska: I found it cute too! But I agree it is not necessary!

thomaska: Maybe `failed` though is not that accurate?! We may fail to send an email because all the instances of an owner are new, or because the "owner" doesn't own any instances!

app/services/notification/Notifications.scala
thomaska: I will integrate the 2 strings used from `Configuration` into `amiableConfigProvider` so that only that is passed around!

app/services/notification/Notifications.scala
thomaska: Yes sorry you had commented about them also in the past and I didn't respond... I just found them here:
https://github.com/guardian/amigo/blob/f172f42c0732385b11da22de9540b12699d2fe49/app/schedule/BakeScheduler.scala#L48
and thought to keep it the same... I will remove them.

app/services/notification/AWSMailClient.scala
thomaska: This must be a result of optimise imports...I 'll amend!

app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -0,0 +1,49 @@
...
+  // Message id used if an owner doesn't have any old instances, or doesn't own any instances at all
+  val MessageNotSent = "0"
+
+  def run(mailClient: AWSMailClient)(implicit config: AMIableConfig, ec: ExecutionContext): Attempt[List[String]] = {
+    for {
+      instancesWithAmis <- Prism.instancesWithAmis(SSA(stage = Some("PROD")))
+      oldInstances = PrismLogic.oldInstances(instancesWithAmis)
+      owners <- Prism.getOwners
+      mailIds <- Attempt.traverse(owners) { owner =>
+        val ownerOldInstances = instancesForOwner(owner, oldInstances)
```
sihil: Doesn't this have to be the other way round in order to avoid notifying two users about the same instance? i.e. shouldn't this look at each instance individually, find the most appropriate owner for an instance, then group by owner and send out the e-mails.

sihil: Just noticed from the PR description that this is your intention, so perhaps this isn't relevant.

thomaska: :/ Yes this is what I though intuitively, but maybe only one person should be notified?! Happy to amend if this is the case!

adamnfish:
:+1:

thomaska:
@sihil Pushed the changes we did, would you be happy to merge!?

PR #44, opened 2017-04-20
Additional attempt type functionality

adamnfish:
Use case from #43.

Applies some Red Book learnings to the Attempt type. It'd be worth considering switching this for EitherT[Future, AmiableErrors, A] but it might not end up heaps simpler. There are some bespoke helpers on here that make this one very convenient to use in this project.

Added a couple of tests for functionality less well covered by the types.

### DIFF
--- a/app/models/Attempt.scala
+++ b/app/models/Attempt.scala
@@ -20,10 +20,10 @@ case class Attempt[A] private (underlying: Future[Either[AMIableErrors, A]]) {
     asFuture.map(_.fold(failure, success))
   }
 
-  def map2[B, C](a2: Attempt[B])(f: (A, B) => C)(implicit ec: ExecutionContext): Attempt[C] = {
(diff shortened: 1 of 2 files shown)


app/models/Attempt.scala
```diff
@@ -79,9 +88,7 @@ object Attempt {
     */
   def successfulAttempts[A](attempts: List[Attempt[A]])(implicit ec: ExecutionContext): Attempt[List[A]] = {
     Attempt.Async.Right {
-      Future.sequence(attempts.map { attempt =>
-        attempt.fold(_ => None, a => Some(a))
-      }).map(_.flatten)
+      Future.traverse(attempts)(_.fold(_ => None, a => Some(a))).map(_.flatten)
```
TBonnin: That might be just me but I found this line quite hard to read with all those parenthesis. 🙈

adamnfish: Great feedback, thank you!

PR #45, opened 2017-05-11
Add usage button to AMI modal overlay

jennysivapalan:
I think this will make it easier to find where an AMI is being used. Button links to Usage section on`/ami?imageId=[amiId]`. 

picture 59

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,5 +35,5 @@ For this reason, in order to set up locally, the following steps must be followe
  In order to setup the auth parameters (eg. `serviceAccountCertPath`),
  please consult someone from the Dev Tools team.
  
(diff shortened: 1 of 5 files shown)


sihil:
Great idea. As Thomas says, it's unclear which AMI is being linked to. Perhaps a "- view usage" after each AMI id?

jennysivapalan:
Ahh, I hadn't thought of that (I always assume every app is using the old AMI). Yes that makes sense, I'll make that change.

jennysivapalan:
@thomaska @sihil Done!
picture 60

sihil:
Looks reasonable. Despite having suggested it I'm slightly uneasy at mixing information with links that don't look like buttons. Having said that, this is a useful feature that I've wanted a few times. 👍

PR #46, opened 2017-05-16
Replace runtime DI with compile time DI

thomaska:
This is an attempt to replace runtime DI with compile time DI.
The benefits of this change are:
- More consistency with the other Play apps in the department
- Better performance in the application startup
- Less magic - things missing can be checked at compile time.

Disadvantages:
- Having to test all the application features, if anything was missed during the transition.
- The effort of actually doing/maintaining this.

The app has been tested locally, but not all of its features (i.e. sending metrics).


app/AppLoader.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def amiableConfigProvider = new AmiableConfigProvider(wsClient, configuration)
+
+  def amiableController = new AMIable(amiableConfigProvider, agents)(defaultContext)
+
+  def healthCheckController = new Healthcheck
+
+  def loginController = new Login(amiableConfigProvider, wsClient)(defaultContext)
+
+  def assets = new Assets(httpErrorHandler)
```
TBonnin: Use `lazy val` rather than `def` to ensure only one instance can be created

marialivia16:
This PR is very old, not sure if we want to go forward with this change. Either way, I rebased against master because what better way there is to get familiar with a project than resolving merging conflicts? 😄

TBonnin:
It was not far from being completed if I remember correctly. That would be a shame not to do the extra effort to merge this. I am happy to help you @marialivia16 if you want to spend some time on it. 
I don't think there is much left other than taking into account @adamnfish's comments

thomaska:
@marialivia16 as far as I remember this PR was opened before merging the "team notifications" functionality and the only pending thing was to rebase/apply the changes from that PR(s)!

marialivia16:
@TBonnin @thomaska I'm happy to pick this up. Adam's comments seem to have been addressed already. I could use some help setting this up locally whenever you have some time. Thanks!

thomaska:
Yeah @marialivia16 you would need to setup nginx locally... @TBonnin had provided me the stuff to put into `$HOME/.gu/amiable.local.conf`... and don't forget to run it using the local sbt file! (ie. `./sbt run`) !

marialivia16:
With @TBonnin 's help I tested this locally and it looks fine. Is everybody happy with merging this?

@adamnfish

adamnfish:
Changes look fine to me so if you've tested it then :+1:

Thanks heaps for looking at this!

PR #47, opened 2017-05-24
Fixes in team notifications functionality

thomaska:
- Added an override to email address in configuration
- Added an endpoint as a  manual notification trigger
- Tweaks in the notification email
- Added restrictions for sending notifications in Dev mode

### DIFF
--- a/app/services/notification/ScheduledNotificationRunner.scala
+++ b/app/services/notification/ScheduledNotificationRunner.scala
@@ -33,15 +33,7 @@ class ScheduledNotificationRunner@Inject()(mailClient: AWSMailClient, environmen
   }
 
   def findInstanceOwners(instances: List[Instance], owners: Owners): Map[Owner, List[Instance]] = {
-    instances.map { i => (i, ownerForInstance(i, owners)) }.groupBy(_._2).mapValues(_.map(_._1))
-  }
-
-  def ownerForInstance(i: Instance, owners: Owners): Owner = {
-    owners.owners.find(_.hasSSA(SSA(i.stack, i.stage, i.app.headOption)))
-      .orElse(owners.owners.find(_.hasSSA(SSA(i.stack, app = i.app.headOption))))
-      .orElse(owners.owners.find(_.hasSSA(SSA(i.stack, i.stage))))
-      .orElse(owners.owners.find(_.hasSSA(SSA(i.stack))))
-      .getOrElse(owners.defaultOwner)
+    instances.map { i => (i, ScheduledNotificationRunner.ownerForInstance(i, owners)) }.groupBy(_._2).mapValues(_.map(_._1))
   }
 
   private def createEmailRequest(owner: Owner, instances: Seq[Instance], config: AMIableConfig) = {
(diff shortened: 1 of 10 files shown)


app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -1,22 +1,33 @@
...
       instancesWithAmis <- Prism.instancesWithAmis(SSA(stage = Some("PROD")))
       oldInstances = PrismLogic.oldInstances(instancesWithAmis)
       ownersWithDefault <- Prism.getOwners
       ownersAndOldInstances = findInstanceOwners(oldInstances, ownersWithDefault)
       mailIds <- Attempt.traverse(ownersAndOldInstances.toList) { case (owner, oldInstancesForOwner) =>
-        val request = createEmailRequest(owner, oldInstancesForOwner)
-        mailClient.send(owner, request)
+        val instances = oldInstancesForOwner.sortBy(i => (i.stack, i.stage, i.app.headOption))
+        val request = createEmailRequest(owner, instances, config)
+        if(environment.mode == Mode.Prod || config.overrideToAddress.nonEmpty)
```
thomaska: Thought to leave the logic here as otherwise we would need to have a global override address which might be confusing in other use cases

sihil: Let's explicitly add in the braces for the block.

sihil: Or... how does something like this feel? It moves all of the development / override logic into a single location. It would be ideal if this could be extracted into the companion object and have some test coverage (unfortunately I think you might need to mock the mail client).

```scala
(environment.mode, config.overrideToAddress) match {
  case (Mode.Prod, maybeOverride) =>
    mailClient.send(maybeOverride.getOrElse(s"${owner.id}@guardian.co.uk"), owner, request)
  case (_, Some(override)) =>
    mailClient.send(override, owner, request)
  case (_, None) =>
    Logger.info(s"Not in Prod and no override To Address set. Would have sent email to ${owner.id}, request: $request")
    Attempt.Right("")
}
```

Maybe @adamnfish has some better ideas on how to do this.

app/services/notification/ScheduledNotificationRunner.scala
```diff
@@ -1,40 +1,44 @@
...
-      .orElse(owners.owners.find(_.hasSSA(SSA(i.stack, app = i.app.headOption))))
-      .orElse(owners.owners.find(_.hasSSA(SSA(i.stack, i.stage))))
-      .orElse(owners.owners.find(_.hasSSA(SSA(i.stack))))
-      .getOrElse(owners.defaultOwner)
+    instances.map { i => (i, ScheduledNotificationRunner.ownerForInstance(i, owners)) }.groupBy(_._2).mapValues(_.map(_._1))
   }
 
-  private def createEmailRequest(owner: Owner, instances: Seq[Instance])(implicit config: AMIableConfig) = {
-    val destination = new Destination().withToAddresses(s"${owner.id}@guardian.co.uk")
+  private def createEmailRequest(owner: Owner, instances: Seq[Instance], config: AMIableConfig) = {
```
sihil: This looks like it could be part of the companion object as well. Also worth adding tests for the behaviour of the overrideToAddress.

PR #50, opened 2018-01-11
Add 'redeploy now' button to amiable which links to riffraff.

philmcmahon:
This should speed up the process of updating out of date AMIs.

It looks like this:
screen shot 2018-01-11 at 17 18 49

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,3 +37,6 @@ For this reason, in order to set up locally, the following steps must be followe
  
  - `./sbt run` open your browser at `https://amiable.local.dev-gutools.co.uk`!
  
+### Common problems
+ - If when running master you can an error "Could not find a suitable constructor..." it's something wrong with your
+ config file - you probably need to add include "application.conf" to you application.local.conf file.
--- a/README.md
+++ b/README.md
@@ -39,4 +39,4 @@ For this reason, in order to set up locally, the following steps must be followe
  
 ### Common problems
  - If when running master you can an error "Could not find a suitable constructor..." it's something wrong with your
- config file - you probably need to add include "application.conf" to you application.local.conf file.
+ config file - you probably need to add `include "application.conf"` to you application.local.conf file.
--- a/app/models/models.scala
+++ b/app/models/models.scala
@@ -84,6 +84,13 @@ case class SSA (
   app: Option[String] = None
 ) {
   def isEmpty = stack.isEmpty && stage.isEmpty && app.isEmpty
+  def riffRaffLink: Option[String] = for {
+    stackName <- stack
+    stageName <- stage
+    appName <- app
+  } yield {
+    s"https://riffraff.gutools.co.uk/deployment/target/deploy?region=eu-west-1&stack=$stackName&stage=$stageName&app=$appName"
+  }
   override def toString: String = s"SSA<${stack.getOrElse("none")}, ${stage.getOrElse("none")}, ${app.getOrElse("none")}>"
 }
 object SSA {
--- a/app/models/models.scala
+++ b/app/models/models.scala
@@ -84,13 +84,6 @@ case class SSA (
   app: Option[String] = None
 ) {
   def isEmpty = stack.isEmpty && stage.isEmpty && app.isEmpty
-  def riffRaffLink: Option[String] = for {
-    stackName <- stack
-    stageName <- stage
-    appName <- app
-  } yield {
-    s"https://riffraff.gutools.co.uk/deployment/target/deploy?region=eu-west-1&stack=$stackName&stage=$stageName&app=$appName"
-  }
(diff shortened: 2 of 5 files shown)


nginx/amiable.sh
```diff
@@ -5,15 +5,7 @@
...
-SYSTEM=$(uname -s)
-if [ $SYSTEM == "Darwin" ]; then
-    # Mac OS X platform
-    confDir=servers
-elif [ $SYSTEM == "Linux" ]; then
-    confDir=sites-enabled
-fi
-
-sudo ln -fs $DIR/amiable.conf $nginxHome/$confDir/amiable.conf
+sudo ln -fs $DIR/amiable.conf $nginxHome/sites-enabled/amiable.conf
```
mchv: Why removing that? On my instance with `homebrew`, `servers` directory is included in the default configuration instead of `sites-enabled`, i.e
```
include servers/*;
```

philmcmahon: hmm ok. On my mac its called sites-enabled which meant this didn't work. I'll put it back or try and make it smarter!

nginx/amiable.sh
```diff
@@ -5,7 +5,13 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 nginxHome=`nginx -V 2>&1 | grep "configure arguments:" | sed 's/[^*]*conf-path=\([^ ]*\)\/nginx\.conf.*/\1/g'`
 
-sudo ln -fs $DIR/amiable.conf $nginxHome/sites-enabled/amiable.conf
+if [ -d "${nginxHome}/sites-enabled" ]; then
+  confDir=sites-enabled
+else
+  confDir=servers
+fi
```
mchv: Unfortunately it will still not work as the directory will exist, it is just that the in the main configuration file, `servers` is  included instead of `sites-enabled`. I think It may be easier to tell people to change that line if there are using `OS x`. I know `https://github.com/guardian/dev-nginx` has the same issue.

README.md
```diff
@@ -36,4 +36,7 @@ For this reason, in order to set up locally, the following steps must be followe
  please consult someone from the Dev Tools team.
  
  - `./sbt run` open your browser at `https://amiable.local.dev-gutools.co.uk`!
- 
\ No newline at end of file
+ 
+### Common problems
+ - If when running master you can an error "Could not find a suitable constructor..." it's something wrong with your
+ config file - you probably need to add include "application.conf" to you application.local.conf file.
```
adamnfish: `include "application.conf"` (with backticks) would be clearer here.

app/models/models.scala
```diff
@@ -84,6 +84,13 @@ case class SSA (
   app: Option[String] = None
 ) {
   def isEmpty = stack.isEmpty && stage.isEmpty && app.isEmpty
+  def riffRaffLink: Option[String] = for {
+    stackName <- stack
+    stageName <- stage
+    appName <- app
+  } yield {
+    s"https://riffraff.gutools.co.uk/deployment/target/deploy?region=eu-west-1&stack=$stackName&stage=$stageName&app=$appName"
```
adamnfish: Why is the region hard-coded?

adamnfish: I also feel like the inclusion of this function on the model is borderline (leaning towards it being wrong).

Models are your datastructures. The rule of thumbs I use for including things on the models are:

* generally don't
* if it is an inherent property of the data structure and it entirely emerges from the information that the model contains then that probably makes sense to be included
* sometimes you really have to (e.g. `toString`)

In particular this means it mustn't take any arguments. While this function doesn't take any args and thus looks ok on the face of it, that's only because the Riff Raff URL is hard coded. That information isn't wholly contained in the data structure so I'd consider having this be a function that takes an SSA.

adamnfish: The above is only a mater of taste though, so feel free to ignore me / come chat :-)

adamnfish: p.s. the page this links to is very useful!

app/views/fragments/amiOverlay.scala.html
```diff
@@ -67,6 +67,13 @@
             </div>
         }
         <a href="#!" class="black-text modal-action modal-close waves-effect waves-white btn grey lighten-3">Close</a>
+        @if(ami.upgrade.isDefined) {
+            @stackStageApp.map { ssa =>
+                @ssa.riffRaffLink.map { link =>
+                    <a href="@link" target="_blank" rel="noopener" class="waves-effect waves-light btn">Redeploy now</a>
```
adamnfish: I think just `Redeploy` would be clearer. `Redeploy now` makes it sound more like that button actually kicks off a deployment and is more verbose than required.

PR #55, opened 2020-05-06
add a link to riffraff next to SSA

akash1810:
There are two main reasons why amiable will flag an app as running an old instance:
- AMIgo hasn't built an image recently
- RiffRaff hasn't deployed the app recently

Having a link to RiffRaff in the SSA table makes it easier to determine which of the two is the reason, leading to a faster resolution.

# Before
image

image

# After
image

image


akash1810:
Requires https://github.com/guardian/riff-raff/pull/584.

akash1810:
cc @guardian/deploy-infrastructure

PR #56, opened 2020-12-01
Replace hard coded date serialiser with play version.

philmcmahon:
## What does this change?
Following from https://github.com/guardian/prism/pull/99/files#diff-99e8aff9e82352e9e7be6d0002a173dd26ee897fb56621045a0286cc3e99a742R51 the formats of dates in Prism has changed. This PR solves a problem with amiable where it was failing to parse the new date format coming from prism for the `createdAt` datetime on the `Instance` object.

The solution replaces manual date formatters with ones comng from the play library, which seems a sensible improvement (thanks @sihil )

## How to test

Run amiable, check that the UI doesn't throw a big error like this one:

Screenshot 2020-12-01 at 14 46 33

### DIFF
--- a/app/models/models.scala
+++ b/app/models/models.scala
@@ -3,6 +3,8 @@ package models
 import org.joda.time.DateTime
 import play.api.libs.json._
 import utils.{DateUtils, Percentiles}
+import play.api.libs.json.JodaWrites._
+import play.api.libs.json.JodaReads._
 
 case class AMI(
   arn: String,
--- a/test/util/Fixtures.scala
+++ b/test/util/Fixtures.scala
@@ -67,7 +67,7 @@ object Fixtures {
                           |      "ip": "0.0.0.0"
                           |    }
                           |  },
-                          |  "createdAt": "2016-02-09T10:59:01.000Z",
+                          |  "createdAt": "2020-11-13T18:56:59Z",
                           |  "instanceName": "i-id",
                           |  "region": "region",
                           |  "vendor": "aws",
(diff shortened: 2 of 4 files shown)


test/util/Fixtures.scala
```diff
@@ -67,7 +67,7 @@ object Fixtures {
                           |      "ip": "0.0.0.0"
                           |    }
                           |  },
-                          |  "createdAt": "2016-02-09T10:59:01.000Z",
+                          |  "createdAt": "2020-11-13T18:56:59Z",
```
adamnfish: Presumably it's not just this one and we'd need to look at the other dates in here as well (e.g. the AMI)? Depends what we decide with the Prism change though. Ideally the Prism format is now stable so that we don't need to chase up fixes in other services? Let's see what DevX decide.

adamnfish (approved):
This looks good, we can decide separately about the fixtures based on what gets decided for the stability of Prism's JSON representation.

test/util/Fixtures.scala
adamnfish: Sorry, commented here before you removed it :-)

philmcmahon: ah yeah good shout I was being so lazy I didn't check the other fixtures!

app/models/models.scala
```diff
@@ -3,6 +3,8 @@ package models
 import org.joda.time.DateTime
 import play.api.libs.json._
 import utils.{DateUtils, Percentiles}
+import play.api.libs.json.JodaWrites._
```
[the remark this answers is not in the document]
philmcmahon: cool, amiable is in theory about to get a load of dev time so I can bear this in mind

PR #60, opened 2021-04-14
Add metric for total no. old dinstances

philmcmahon:
## What does this change?
This PR
 - Changes the namespace of all custom cloudwatch metrics from amiable to include the Stage so that we can distinguish between PROD and CODE amiable (previously there was no amiable CODE stack)
 - Adds a new 'total old instances' metric with 'account' dimension so it can be broken down based off AWS account.

## How to test
This is on CODE at the moment and seems to be working:
Screenshot 2021-04-14 at 11 05 47

A useful metric in the [grafana vulnerabilities dashboard](https://metrics.gutools.co.uk/d/RHAvkh8Gz/devx-securityhq?orgId=1)

### DIFF
--- a/app/metrics/CloudWatch.scala
+++ b/app/metrics/CloudWatch.scala
@@ -28,7 +28,7 @@ object CloudWatchMetrics {
   case object OldCountByAccount extends CloudWatchMetric("instances-running-out-of-date-amis-account")
 }
 
-class CloudWatch(amiableConfigProvider: AmiableConfigProvider) {
+class CloudWatch(stage: String) {
   lazy val client = {
     val credentialsProvider = new AWSCredentialsProviderChain(
       InstanceProfileCredentialsProvider.getInstance(),
@@ -41,7 +41,7 @@ class CloudWatch(amiableConfigProvider: AmiableConfigProvider) {
     acwac
   }
 
-  val namespace = s"AMIable-${amiableConfigProvider.stage}"
+  val namespace = s"AMIable-$stage"
 
   private[metrics] def putRequest(metricName: String, value: Int, dimensions: List[Dimension] = List.empty): PutMetricDataRequest = {
     new PutMetricDataRequest()
--- a/app/metrics/CloudWatch.scala
+++ b/app/metrics/CloudWatch.scala
@@ -41,11 +41,11 @@ class CloudWatch(stage: String) {
     acwac
   }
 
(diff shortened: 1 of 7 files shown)


jorgeazevedo:
Are you opposed of sending this to the SecurityHQ namespace? Say we're doing a sum of all the vulnerabilities for identity. We currently need to do a query that looks like this

```
SUM(SEARCH('{SecurityHQ,Account,DataType} MetricName=Vulnerabilities AND Account=Identity AND (DataType="iam/credentials/total" OR DataType="iam/keys/total" OR DataType="securitygroup/total")', 'Average', 900))
```

If the metric is not in the same namespace, I'm not sure how to make it work exactly. The most natural way is for it to live in the same namespace, have the same MetricName, and just become a new OR clause.

On the DataType, currently we have

```
iam/credentials/total
iam/keys/total
iam/keys/critical
iam/keys/warning
securitygroup/total
```

Would you be opposed to this one being something slash total? like `oldami/total`? It'd be neater, but it might bring the benefit of simplifying the query to be all "total" datatypes though I haven't tried it yet

philmcmahon:
@jorgeazevedo that makes sense, I've updated the namespace for this metric and the dimension in https://github.com/guardian/amiable/pull/60/commits/49eabe510c480e11ca77e959bb02a74693979586

jorgeazevedo (approved):
🚀

jorgeazevedo:
thanks! can't wait to get this data in

app/metrics/CloudWatch.scala
```diff
@@ -37,22 +41,11 @@ object CloudWatch {
...
-      .withValue(prodStage),
-    new Dimension()
-      .withName("stack")
-      .withValue(allStacks)
-  )
-
-  private[metrics] def putRequest(metricName: String, value: Int): PutMetricDataRequest = {
+  private[metrics] def putRequest(metricName: String, value: Int, dimensions: List[Dimension] = List.empty, namespace: String = defaultNamespace): PutMetricDataRequest = {
     new PutMetricDataRequest()
-      .withNamespace(namespace)
```
jorgeazevedo, line 55: @philmcmahon I noticed that the PROD metrics are still coming through to the AMIable-PROD namespace. Am I reading this right, is putRequest ignoring the namespace argument and just using defaultNamespace? Which would justify what's happening?

PR #63, opened 2021-04-23
Stop sending metrics from code

jorgeazevedo:
## What does this change?

We have a problem with using the `SecurityHQ` namespace for one of these metrics. There's two namespaces currently for amiable, `amiable-PROD` and `amiable-code`, and each environment posts metrics to their corresponding namespace. This is defined in `CloudWatch.scala` in the constructor. And you can see the problem - by forcing the namespace to `SecurityHQ`, we get both environments sending metrics there, which is not desired behaviour.

Except... `Metrics.scala` has a big if guard clause up top that says `// only add metrics from PROD`, so in theory there shouldn't be any metrics being sent from amiable code.

So the guard clause isn't working. I've patched to use the same data source as CloudWatch to decide if it's CODE or PROD and it seems to work. CODE has stopped posting metrics.

Why are `amiableConfigProvider.stage` and `environment.mode` giving different results? Perhaps fixing that would be a better patch than what I'm proposing here?

### DIFF
--- a/app/AppLoader.scala
+++ b/app/AppLoader.scala
@@ -50,7 +50,7 @@ class AppComponents(context: Context) extends play.api.BuiltInComponentsFromCont
   val cloudwatch = new CloudWatch(amiableConfigProvider.stage)
 
   val agents = new Agents(amiableConfigProvider, applicationLifecycle, actorSystem, environment, cloudwatch)
-  val metrics = new Metrics(cloudwatch, environment, agents, applicationLifecycle)
+  val metrics = new Metrics(cloudwatch, amiableConfigProvider.stage, agents, applicationLifecycle)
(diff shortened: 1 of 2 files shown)


akash1810:
> Why are `amiableConfigProvider.stage` and `environment.mode` giving different results? Perhaps fixing that would be a better patch than what I'm proposing here?

I believe `environment.mode` represents Play's running mode, rather than our concept of `Stage`.

akash1810 (approved):

akash1810:
We should add a similar guard to the email notifications - only PROD should send emails on a cron.

jorgeazevedo:
thanks @akash1810 , I'd forgotten about this PR!

PR #65, opened 2021-05-27
fix: Only setup scheduled notifications in PROD

akash1810:
Easiest to review [w/out whitespace](https://github.com/guardian/amiable/pull/65/files?diff=split&w=1).

## What does this change?

We're currently duplicating scheduled emails (once per stage).

This makes inboxes quite noisy.

This change reduces the noise by running the schedule in PROD only.

If we want to execute the sending of emails in CODE, we can do this manually via [`POST /sendEmail`](https://github.com/guardian/amiable/blob/8b0ecfc596eb76a29fe984076c57ab38970c2c33/conf/routes#L21).


akash1810:
cc @guardian/developer-experience

philmcmahon (approved):

PR #67, opened 2021-06-11
Always include every account, even if it's zero

jorgeazevedo:
## What does this change?
Report metrics for every known account, even if the number of old instances is zero.

This makes it easier to plot the data.

## How to test
Ran this locally!

### DIFF
--- a/app/services/Agents.scala
+++ b/app/services/Agents.scala
@@ -108,10 +108,11 @@ class Agents @Inject() (amiableConfigProvider: AmiableConfigProvider, lifecycle:
       accounts <- Prism.getAccounts
     } yield {
       val now = DateTime.now
-      val amisForAccount = instancesWithAmis.groupBy(_._1.meta.origin.accountName.getOrElse("unknown-account"))
+      val oldInstancesForAccount = PrismLogic.oldInstances(instancesWithAmis).groupBy(_.meta.origin.accountName.getOrElse("unknown-account"))
 
-      val oldInstanceCountsByAccount = accounts.map( account => {
-          OldInstanceAccountHistory(now, account.accountName, PrismLogic.oldInstances(amisForAccount.getOrElse(account.accountName, List())).length)
+      val oldInstanceCountsByAccount = accounts.map(account => {
+        val numberOfOldInstancesForAccount = oldInstancesForAccount.getOrElse(account.accountName, List()).length
+        OldInstanceAccountHistory(now, account.accountName, numberOfOldInstancesForAccount)
       })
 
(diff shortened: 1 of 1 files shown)


app/services/Agents.scala
```diff
@@ -104,12 +104,21 @@ class Agents @Inject() (amiableConfigProvider: AmiableConfigProvider, lifecycle:
...
-      .map{ case (accountName, list) => OldInstanceAccountHistory(now, accountName, list.length)}
-      .toList
-    oldInstanceCountByAccountAgent.send(oldInstanceCountsByAccount)
+    for {
+      accounts <- Prism.getAccounts
+    } yield {
+      val now = DateTime.now
+      val amisForAccount = instancesWithAmis.groupBy(_._1.meta.origin.accountName.getOrElse("unknown-account"))
+
+      val oldInstanceCountsByAccount = accounts.map( account => {
```
philmcmahon, line 113: Another way of writing this - though not massively tidier
```
accounts.map { account =>
   val oldInstanceCount = amisForAccount.get(accoount.accountName).map(PrismLogic.oldInstances).getOrElse(0)
OldInstanceAccountHistory(now, account.accountName, oldInstanceCount)
}
```

philmcmahon: actually maybe even cleaner you could call PrismLogic on instancesWithAmis before the group by as was done before, not certain if that  would work though!

app/services/Agents.scala
```diff
@@ -104,12 +104,21 @@ class Agents @Inject() (amiableConfigProvider: AmiableConfigProvider, lifecycle:
...
 
   def refreshOldInstanceCountInfo(instancesWithAmis: List[(Instance, Option[AMI])]): Unit = {
-    val now = DateTime.now
-    val oldInstanceCountsByAccount = PrismLogic.oldInstances(instancesWithAmis)
-      .groupBy(_.meta.origin.accountName.getOrElse("unknown-account"))
-      .map{ case (accountName, list) => OldInstanceAccountHistory(now, accountName, list.length)}
-      .toList
-    oldInstanceCountByAccountAgent.send(oldInstanceCountsByAccount)
+    for {
+      accounts <- Prism.getAccounts
```
philmcmahon, line 108: sometimes when there's only one thing in a for, yield I just do a `.flatMap (so Prism.getAccounts.flatMap(accounts=>` but that might noot be possible on an `Attempt`

philmcmahon (approved):
Nice work

app/services/Agents.scala
jorgeazevedo: It does work! Good shout

PR #72, opened 2021-08-10
chore: Simplify the CI script

akash1810:
## What does this change?

It appears the current build script has some work arounds from the time we were in data centers.

This change simplifies the build for today's setup.

It's worth noting that with this change, the RiffRaff project name goes from "amiable" to "tools::amiable"; I'll add a restriction to the "amiable" project and point people to the updated name.

## How to test

- CI should succeed
- RiffRaff should be able to deploy to CODE

A simpler CI setup.

This change requires a change to the TeamCity project and as a result other branches need to be rebased with these changes to continue to build.


build.sbt
```diff
@@ -50,22 +50,13 @@ resolvers += "scalaz-bintray" at "http://dl.bintray.com/scalaz/releases"
...
 
 riffRaffPackageType := (packageBin in Debian).value
-
-def env(key: String): Option[String] = Option(System.getenv(key))
-
-riffRaffBuildIdentifier := env("BUILD_NUMBER").getOrElse("DEV")
-riffRaffManifestBranch := env("BRANCH_NAME").getOrElse("unknown_branch")
-riffRaffManifestVcsUrl := "git@github.com:guardian/amiable.git"
 riffRaffUploadArtifactBucket := Some("riffraff-artifact")
 riffRaffUploadManifestBucket := Some("riffraff-builds")
```
sihil, line 61: These buckets can also be deleted when using `riffRaffNotifyTeamcity`.

sihil (approved):
Minor comment, but otherwise a nice bit of polishing 🥇

PR #73, opened 2021-08-10
feat: Co-locate infrastructure for CD benefits

akash1810:
## What does this change?

This change adds the CDK defined infrastructure into this repository to enable it to be deployed alongside the app with RiffRaff.

This initially started as a copy/paste from the deploy-tools-platform repository, however I ended up adding a few hygiene tasks too to make it easier to work with. It might be easier to review commit by commit.

## How to test

- Deploy to CODE and see it work? ([it has!](https://riffraff.gutools.co.uk/deployment/view/9085d904-05ad-4740-9140-1d01a08924d2))

No more manual updating of a Cloudformation stack!


riff-raff.yaml
```diff
@@ -19,3 +19,6 @@ deployments:
         AmigoStage: PROD
         BuiltBy: amigo
       amiEncrypted: true
+      templatePath: Amiable.template.json
+      cloudFormationStackName: amiable-gucdk-
```
jacobwinch, line 23: I think we can allow Riff-Raff to discover the stack based on tags (and omit `cloudFormationStackName` and `appendStageToCloudFormationStackName`)? I think this is what we're doing for [Prism](https://github.com/guardian/prism/blob/47f1630a266236c155b9294626973fbb3088df2d/riff-raff.yaml#L4-L14).

jacobwinch (approved):

riff-raff.yaml
akash1810: Ah! That's nicer! Will do this in another PR.

akash1810: I think Riff-Raff is already doing this, which is more reason to remove this property as it appears to be entirely unused.

image

akash1810: https://github.com/guardian/amiable/pull/77

PR #109, opened 2022-01-07
Update amiable to use scala 2.13 along with dependency updates and required code changes

michaelwmcnamara:
## What does this change?
This updates the version of scala used by Amiable from 2.11 to 2.13.
As part of this, other required updates are also applied:

- SBT is updated to the latest version available (1.5.7) This version also incorporates patches to guard against the Log4J remote execution vulnerabillity.
- com.gu.play-googleauth has been updated to the latest version (2.2.2)
-scalatest has been updated to the latest version (3.2.10)

A number of other dependencies and plugins have also been updated to versions that work for scala 2.13.
Code changes:

Significant code changes necessitated by this update include:

- play.api.Logging is now added to objects as a trait, with logger already instantiated as a variable. This changes the way Logger is called.
- Methods called with no parameters must not be called with ()  appended. 
- mapValues is deprecated and must be replaced with view.mapValues

ScalaTest changes:
- Scalatest traits have been split into their own modules. This requires some changes to the import commands used.
- FreeSpec has been renamed to AnyFreeSpec and moved to org.scalatest.freespec.AnyFreeSpec.
- Matchers have been moved to org.scalatest.matchers.should.Matchers.
- MockitoSugar has been moved to org.scalatestplus.mockito.MockitoSugar.

Related Trello card: https://trello.com/c/8YUJgAlx/1030-update-amiable-to-use-sbt-157

## How to test

- [x] Code compiles
- [x] Test runs successfullly on sbt
- [x] Code runs locally
- [x] Deploys and runs successfully in CODE
 

Deploy branch to CODE 
Ensure deploy is successful
Ensure that https://amiable.code.dev-gutools.co.uk/ works correctly

Exposure to vulnerabilities is reduced. If there are new 0 day vulnerabilities that we need to patch, we are able to quickly adapt. Easy to innovate as barrier to adopting new features is smaller.

Only real risks are that the code is broken and so Amiable stops working when we deploy. Testing will guard against this.


build.sbt
```diff
@@ -1,10 +1,12 @@
+import com.typesafe.sbt.packager.debian.DebianPlugin.autoImport.Debian
+
 name := """amiable"""
 
-version := "1.0-SNAPSHOT"
+version := "1.2-SNAPSHOT"
```
akash1810, line 5: Is this change needed?

akash1810 (commented):
Cool! Has this been checked on CODE?

michaelwmcnamara:
Deployed and ran successfully on CODE

build.sbt
michaelwmcnamara: Good spot! Have reverted version back to "1.0-SNAPSHOT"

akash1810 (approved):
Fantastic! Great work! 🎉

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.15
+sbt.version=1.5.7
```
mchv, line 1: I would suggest `1.6.1`

michaelwmcnamara: Great idea @mchv! I have now updated to 1.6.1.

PR #133, opened 2022-05-06
Change log format to JSON

akash1810:
## What does this change?

Change logging format to output as JSON. Combined with https://github.com/guardian/devx-logs/pull/16, this means we'll get structured logging in Central ELK 🎉 .

The Jackson changes are not ideal. Without these changes we get:

```console
[error] com.fasterxml.jackson.databind.JsonMappingException: Scala module 2.11.4 requires Jackson Databind version >= 2.11.0 and < 2.12.0
```

Following the advice from https://github.com/orgs/playframework/discussions/11222 was the only solution we could find. We'll attempt to find a better solution in a future PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,6 +32,27 @@ scalacOptions := Seq(
   "-Ywarn-unused"
 )
 
+// https://github.com/orgs/playframework/discussions/11222
+val jacksonVersion         = "2.13.2"
+val jacksonDatabindVersion = "2.13.2.2"
+
+val jacksonOverrides = Seq(
+  "com.fasterxml.jackson.core"     % "jackson-core",
+  "com.fasterxml.jackson.core"     % "jackson-annotations",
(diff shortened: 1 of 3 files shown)


akash1810:
This seems to have broken the charts on the index page:

image

akash1810:
Regarding the graph showing no data, this is related to our [creating metrics only in PROD](https://github.com/guardian/amiable/blob/32950f2dc6b6229965b45881fe0cdfd308e326d9/app/services/Metrics.scala#L13-L14) and [querying a stage dependant namespace](https://github.com/guardian/amiable/blob/32950f2dc6b6229965b45881fe0cdfd308e326d9/app/metrics/CloudWatch.scala#L44). That is, CODE is querying the CODE namespace, which we're not populating.

Two solutions come to mind:
  - All stages _read_ metrics from PROD namespace, and only PROD _writes_
  - All stages _read_ and _write_ from/to their own namespace (increases cost)

jacobwinch (approved):

PR #137, opened 2022-05-09
feat: Allow pre-PROD environments to read PROD's cloudwatch metrics

akash1810:
Likely to be easiest reviewed commit by commit.

## What does this change?

As seen in #133, we're currently only producing cloudwatch metrics in PROD. This means the graph on the index page in pre-PROD environments will be empty.

In this change we let pre-PROD environments read cloudwatch metrics from the PROD namespace, fixing the graph on the index page.

We also push the namespace of our cloudwatch metrics into configuration for clarity and add a `TODO` against a metric that [uses a different namespace](#60), which was a little confusing.

## How to test

When deployed to CODE, we should see data on the index page's graph.

Spotting regressions in pre-PROD becomes easier? This is true at least for #133.

### DIFF
--- a/app/config/AMIableConfig.scala
+++ b/app/config/AMIableConfig.scala
@@ -43,7 +43,18 @@ class AmiableConfigProvider @Inject() (val ws: WSClient, val playConfig: Configu
     playConfig, "stage"
   )
 
-  val cloudwatchReadNamespace: String = s"AMIable-$stage"
+  /**
+    * Cloudwatch metrics are expensive.
+    * Our pre-PROD environments do not get much traffic, so in order to reduce AWS spend, only create metrics in PROD.
+    */
+  val shouldCreateCloudwatchMetrics: Boolean = stage == "PROD"
+
+  /**
+    * If our pre-PROD environments are not creating metrics, have them read PROD's so the app can be fully tested.
+    * @see [[`shouldCreateCloudwatchMetrics`]]
+    */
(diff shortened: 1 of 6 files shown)


akash1810:
Tested on CODE and works as expected:

image

app/config/AMIableConfig.scala
```diff
@@ -43,6 +43,21 @@ class AmiableConfigProvider @Inject() (val ws: WSClient, val playConfig: Configu
     playConfig, "stage"
   )
 
+  /**
+    * Cloudwatch metrics are expensive.
```
jacobwinch, line 47: Although CloudWatch can be an expensive service, Cost Explorer shows that we are currently spending < $5 per month for Amiable, so we could simplify here and just push the metrics in `CODE` too?

jacobwinch (approved):

app/config/AMIableConfig.scala
akash1810: Ah, I didn't actually look at our spend, just assumed it was high! <$5/month isn't that much. Will update to produce metrics from each stage.

It's a strange one though as both CODE and PROD use the same source of data (Prism + AWS accounts), so the metrics will be exactly the same. IIUC each AMIable stage uses Prism PROD too (and in reality Prism CODE === Prism PROD as they scrape the same AWS accounts).

akash1810:
Opting for this over https://github.com/guardian/amiable/pull/142 as we're unlikely to ever use metrics produced from CODE.

PR #142, opened 2022-05-19
feat: Produce CloudWatch Metrics for every stage, not just `PROD`

akash1810:
## What does this change?

A replacement for #137, this change updates AMIable to always produce CloudWatch Metrics (currently metrics are only produced in PROD).

Simpler, more obvious behaviour?

We're currently shipping metrics to two namespaces:
  1. `AMIable-<STAGE>`
  2. `SecurityHQ`

The second namespace will now have twice the number of metrics for the same data with no identity as to what produced it. Will this cause confusion?

### DIFF
--- a/app/services/Metrics.scala
+++ b/app/services/Metrics.scala
@@ -20,6 +20,9 @@ class Metrics(cloudWatch: CloudWatch, stage: String, namespace: String, security
       cloudWatch.put(namespace, CloudWatchMetrics.AmisAgePercentile75th.name, agents.amisAgePercentiles.flatMap(_.p75))
       cloudWatch.put(namespace, CloudWatchMetrics.AmisAgePercentile90th.name, agents.amisAgePercentiles.flatMap(_.p90))
       cloudWatch.put(namespace, CloudWatchMetrics.AmisAgePercentileHighest.name, agents.amisAgePercentiles.flatMap(_.highest))
+
+      // These metrics are used by a Grafana dashboard which graphs SecurityHQ related metrics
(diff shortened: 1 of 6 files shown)


app/services/Metrics.scala
```diff
@@ -2,30 +2,27 @@ package services
...
-      cloudWatch.put(CloudWatchMetrics.AmisAgePercentile50th.name, agents.amisAgePercentiles.flatMap(_.p50))
-      cloudWatch.put(CloudWatchMetrics.AmisAgePercentile75th.name, agents.amisAgePercentiles.flatMap(_.p75))
-      cloudWatch.put(CloudWatchMetrics.AmisAgePercentile90th.name, agents.amisAgePercentiles.flatMap(_.p90))
-      cloudWatch.put(CloudWatchMetrics.AmisAgePercentileHighest.name, agents.amisAgePercentiles.flatMap(_.highest))
-      cloudWatch.put(CloudWatchMetrics.OldCountByAccount.name, agents.oldInstanceCountByAccount)
-    }
+    // These metrics are used by a Grafana dashboard which graphs SecurityHQ related metrics
+    // TODO use the same namespace as other metrics, and filter in Grafana instead?
+    cloudWatch.put(securityHqNamespace, CloudWatchMetrics.OldCountByAccount.name, agents.oldInstanceCountByAccount)
+  }
```
akash1810, line 22: Every stage will produce the same metric on the same namespace. That is there will be duplication. Should we only produce metrics on `securityHqNamespace` in PROD?

kenoir (approved):

PR #143, opened 2022-05-19
Configure CloudWatch alarms for PROD environment and improve test coverage

jacobwinch:
## What does this change?

**Note: this will be easier to review commit by commit.**

This PR configures alarms for the `PROD` environment, as we recently noticed that these were not defined.

When I first made this change I was surprised to see that the snapshot tests passed. This was because we were only running a snapshot test against the `CODE` stack.

Following https://github.com/guardian/cdk/pull/1134, we are now generating two separate templates and it's possible for these to differ, so I think it is important to snapshot test both versions. I have made that change in this PR too (https://github.com/guardian/amiable/pull/143/commits/9e8456e9c3c937c93120ab69d2cd4a4428b7dd9c).

## How to test

Snapshot testing should be sufficient here - this is the relevant commit: https://github.com/guardian/amiable/pull/143/commits/6c4055331d1309d01a89b53a6c0a3035c8b47e82.

We should receive automated alarms for (severe) failures in `PROD`, rather than relying on teams to report issues.

Yes, I think the biggest risk here is accidentally adding excessively noisy alarms; let's try to update these thresholds if necessary.

### DIFF
--- a/cdk/bin/amiable.ts
+++ b/cdk/bin/amiable.ts
@@ -4,18 +4,21 @@ import { App } from "aws-cdk-lib";
 import { Amiable } from "../lib/amiable/amiable";
 
 const app = new App();
-new Amiable(app, "Amiable-CODE", {
-  migratedFromCloudFormation: true,
-  stack: "deploy",
-  stage: "CODE",
-  env: { region: "eu-west-1" },
-  domainName: "amiable.code.dev-gutools.co.uk",
-});
 
-new Amiable(app, "Amiable-PROD", {
-  migratedFromCloudFormation: true,
-  stack: "deploy",
+const commonProps = { migratedFromCloudFormation: true, stack: "deploy", env: { region: "eu-west-1" } };
+
+export const codeProps = {
+  ...commonProps,
+  stage: "CODE",
+  domainName: "amiable.code.dev-gutools.co.uk",
+};
+
+export const prodProps = {
+  ...commonProps,
   stage: "PROD",
-  env: { region: "eu-west-1" },
   domainName: "amiable.gutools.co.uk",
-});
+};
+
+new Amiable(app, "Amiable-CODE", codeProps);
+
+new Amiable(app, "Amiable-PROD", prodProps);
--- a/cdk/lib/amiable/__snapshots__/amiable.test.ts.snap
+++ b/cdk/lib/amiable/__snapshots__/amiable.test.ts.snap
@@ -882,3 +882,886 @@ Object {
   },
 }
 `;
+
+exports[`The Amiable stack matches the snapshot 2`] = `
+Object {
+  "Outputs": Object {
+    "LoadBalancerAmiableDnsName": Object {
+      "Description": "DNS entry for LoadBalancerAmiable",
+      "Value": Object {
+        "Fn::GetAtt": Array [
+          "LoadBalancerAmiable9C4DD4AF",
+          "DNSName",
+        ],
+      },
+    },
+  },
+  "Parameters": Object {
+    "AMIAmiable": Object {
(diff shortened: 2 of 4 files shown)


cdk/bin/amiable.ts
```diff
@@ -4,18 +4,21 @@ import { App } from "aws-cdk-lib";
 import { Amiable } from "../lib/amiable/amiable";
 
 const app = new App();
-new Amiable(app, "Amiable-CODE", {
-  migratedFromCloudFormation: true,
-  stack: "deploy",
+
+const commonProps = { migratedFromCloudFormation: true, stack: "deploy", env: { region: "eu-west-1" } };
+
+export const codeProps = {
```
jacobwinch, line 10: Using `const`s here allows us to pass identical props into the tests. This prevents duplication and makes the snapshot tests as close as possible to the actual templates that will be deployed.

cdk/lib/amiable/__snapshots__/amiable.test.ts.snap
```diff
@@ -882,3 +882,1085 @@ Object {
   },
 }
 `;
+
+exports[`The Amiable stack matches the snapshot 2`] = `
```
jacobwinch, line 886: This is brand new because we weren't testing the `PROD` snapshot at all before. For the changes related to this PR, see: https://github.com/guardian/amiable/commit/6c4055331d1309d01a89b53a6c0a3035c8b47e82.

cdk/lib/amiable/amiable.test.ts
```diff
@@ -1,18 +1,16 @@
...
-    const stack = new Amiable(app, "amiable", {
-      migratedFromCloudFormation: true,
-      stack: "deploy",
-      stage: "CODE",
-      env: { region: "eu-west-1" },
-      domainName: "amiable.code.dev-gutools.co.uk",
-    });
 
-    expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
+    const codeStack = new Amiable(app, "amiable", codeProps);
```
jacobwinch, line 10: Updating `"amiable"` id here results in changes to the snapshot, so I have decided it's better to tolerate this inconsistency 😁

cdk/lib/amiable/amiable.ts
```diff
@@ -5,7 +5,6 @@ import { GuAllowPolicy, GuSESSenderPolicy } from "@guardian/cdk/lib/constructs/i
 import { GuPlayApp } from "@guardian/cdk/lib/patterns/ec2-app";
 import { GuardianPublicNetworks } from "@guardian/private-infrastructure-config";
 import type { App } from "aws-cdk-lib";
-import { Tags } from "aws-cdk-lib";
```
jacobwinch, line 8: I should have removed this in https://github.com/guardian/amiable/pull/138!

cdk/lib/amiable/amiable.ts
```diff
@@ -38,7 +37,17 @@ export class Amiable extends GuStack {
       certificateProps: {
         domainName: props.domainName,
       },
-      monitoringConfiguration: { noMonitoring: true },
+      monitoringConfiguration:
+        this.stage === "PROD"
+          ? {
+              snsTopicName: "devx-alerts",
+              http5xxAlarm: {
+                tolerated5xxPercentage: 99,
```
jacobwinch, line 45: It's really hard to tune alarms for low-traffic services like this. Because we are using % based alerting, it's quite possible that 1 or 2 requests resulting in a 5XX response would lead to a high % of overall requests being 5XXs. Perhaps we need to think about supporting an alarm based on absolute number of requests for services like this...

For now, I have decided to attempt to catch the relatively severe cases only e.g. Amiable is completely down for 2 minutes (or more) or Amiable is up but there are 2 consecutive minutes where all responses are 5XXs.

jfsoul (approved):
Looks great, thanks

PR #151, opened 2022-06-09
Bump instance size to small

jfsoul:
## What does this change?
Moves the instance size from micro to small, because we are seeing some instability. CPU appears to spike before the instance becomes unresponsive and perhaps they are being overwhelmed during the scheduled tasks occasionally. If this does not solve the issue it will also be useful information.


jacobwinch (approved):
The change looks good, but I think you'll need to update the snapshot too!

👍

akash1810:
This is similar to [Janus](https://github.com/guardian/janus/pull/3017). Wonder if there's any correlation?

PR #157, opened 2022-07-15
move cdk and dependencies to devDependencies

michaelwmcnamara:
## What does this change?
This PR moves the cdk library and its related dependencies into devDependencies instead of core dependencies.
This is a better reflection of reality as cdk is used to generate the stack, not to actually run the code.
The later versions of the cdk script set the libraries up in this manner, so this change will bring amiable up to date with cdk best-practice.

## How to test

No changes to operational code. If it deploys and builds it is successful.

Amiable is up to date with our cdk best practices

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -27,13 +27,13 @@
     "prettier": "^2.6.2",
     "ts-jest": "^27.1.4",
     "ts-node": "^10.2.0",
-    "typescript": "~4.4.3"
-  },
-  "dependencies": {
+    "typescript": "~4.4.3",
     "aws-cdk": "2.23.0",
     "aws-cdk-lib": "2.23.0",
     "constructs": "10.1.5",
-    "@guardian/cdk": "43.2.0",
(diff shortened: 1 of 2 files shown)


cdk/package.json
```diff
@@ -27,13 +27,13 @@
...
-  "dependencies": {
+    "typescript": "~4.4.3",
     "aws-cdk": "2.23.0",
     "aws-cdk-lib": "2.23.0",
     "constructs": "10.1.5",
-    "@guardian/cdk": "43.2.0",
+    "@guardian/cdk": "43.2.0"
+  },
+  "dependencies": {
     "source-map-support": "^0.5.16"
```
jfsoul, line 37: Could this dependency be moved to dev as well?

jfsoul (approved):

cdk/package.json
michaelwmcnamara: Not sure - I will take a look and raise in a seperate PR if it can be

PR #419, opened 2023-02-01
Create `amaiable.gutools.co.uk`

akash1810:
## What does this change?

As part of the work to restore pre-incident URLs, this change creates `amiable.gutools.co.uk`. It is a CNAME to the same ALB hosting `public.amiable.gutools.co.uk`.

The TLS certificate used by the ALB is also changed, adding the new domain as a subject alternative domain.

In order to offer seamless DX, we also configure a 301 redirect from the legacy domain to this new one so that requests always land on `amiable.gutools.co.uk`. This is important because the auth config only supports one domain name.

We will remove the legacy domain in a future change.

The above is repeated for the CODE forms too.

Once merged, AMIable will be accessible via `public.amiable.gutools.co.uk` and `amiable.gutools.co.uk`.

> **Note**
> We'd need to change the [config in S3](https://github.com/guardian/amiable/blob/abc13f135d20b75530f842ad52c90dc75a8feed6/cdk/lib/amiable/amiable.ts#L35) to use `amiable.gutools.co.uk` before shipping this change.

### DIFF
--- a/cdk/bin/amiable.ts
+++ b/cdk/bin/amiable.ts
@@ -10,13 +10,13 @@ const commonProps = { migratedFromCloudFormation: true, stack: "deploy", env: { 
 export const codeProps = {
   ...commonProps,
   stage: "CODE",
-  domainName: "public.amiable.code.dev-gutools.co.uk",
+  domainName: "amiable.code.dev-gutools.co.uk",
 };
 
 export const prodProps = {
   ...commonProps,
   stage: "PROD",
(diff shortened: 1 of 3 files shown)


nicl (approved):

jacobwinch (approved):
The infrastructure changes look good to me. 

Reminder that you might also need to update some private config for this app (I had to update this as part of https://github.com/guardian/amiable/pull/389).

PR #479, opened 2023-04-12
Build using GitHub Actions

NovemberTang:
## What does this change?

This PR represents the first part of the CI migration. The application and infrastructure are tested and built, and the packages are uploaded to riff raff as part of the CI script. There is no longer a dependency on TeamCity, meeting the minimum requirements of the ticket.

## How to test

Verify that amiable builds, riff-raff can find it, and it deploys to CODE

Amiable still works as expected, the TeamCity project can be deleted

We are still relying on [sbt-riffraff-artifact](https://github.com/guardian/sbt-riffraff-artifact) to upload the packages to riff-raff. This project is not actively updated so may pose a security risk. A follow-up PR will be raised to remove our dependency on this plugin.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,11 @@ jobs:
       id-token: write
       contents: read
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3.5.1
 
       # Setup AWS credentials to enable uploading to S3 for Riff-Raff.
       # See https://github.com/aws-actions/configure-aws-credentials
-      - uses: aws-actions/configure-aws-credentials@v1
+      - uses: aws-actions/configure-aws-credentials@v2.0.0
         with:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
           aws-region: eu-west-1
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+  CI:
+    runs-on: ubuntu-latest
+
+    # See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
+    permissions:
+      # required by aws-actions/configure-aws-credentials
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v2
```
akash1810, line 22: Let's use [v3](https://github.com/actions/checkout/releases/tag/v3.0.0) here - it's moved to using Node 16, which is an LTS version.

See https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      # Setup AWS credentials to enable uploading to S3 for Riff-Raff.
+      # See https://github.com/aws-actions/configure-aws-credentials
+      - uses: aws-actions/configure-aws-credentials@v1
```
akash1810, line 26: Let's use [v2](https://github.com/aws-actions/configure-aws-credentials/blob/v2.0.0/CHANGELOG.md#200-2023-03-06) here - it's moved to using Node 16, which is an LTS version.

See https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

nicl (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/setup-java@v3
+        with:
+          java-version: '8'
+          distribution: 'adopt'
+          cache: 'sbt'
+
+      # See https://github.com/github/scripts-to-rule-them-all
+      - name: Run script/ci
+        run: |
+          LAST_TEAMCITY_BUILD=1007
```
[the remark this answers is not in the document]
NovemberTang: the teamcity build has already been paused!

PR #481, opened 2023-04-13
build using actions riff-raff

NovemberTang:
## What does this change?

amiable no longer builds using sbt-riffraff-artifact, instead preferring actions-riff-raff

## How to test

verify that you are able to build and deploy to code using this branch

number of vulnerabilities decreases as a result of the migration

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,3 +48,17 @@ jobs:
           LAST_TEAMCITY_BUILD=1007
           export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
           ./script/ci
+
+      - name: Upload to riff-raff
+        uses: guardian/actions-riff-raff@v2
+        with:
+          app: amiable
+          buildNumberOffset: 1007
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -48,3 +48,17 @@ jobs:
           LAST_TEAMCITY_BUILD=1007
           export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
           ./script/ci
+
+      - name: Upload to riff-raff
+        uses: guardian/actions-riff-raff@v2
+        with:
+          app: amiable
+          buildNumberOffset: 1007
+          buildNumber: ${{ env.GITHUB_RUN_NUMBER }}
```
akash1810, line 57: I don't think we need both `buildNumerOffset` and `buildNumber`. It looks like we can [get away with just `buildNumberOffset`](https://github.com/guardian/security-hq/blob/7012d12bb29faf06bdb10690f58f17cd4ec6200d/.github/workflows/ci-sbt.yml#L54-L55).

akash1810 (approved):
Couple of minor comments, but nothing blocking. Great work!

img

PR #514, opened 2023-05-22
build using java 11

NovemberTang:
## What does this change?

A bunch of dependency upgrades will not compile until this project is built using java 11

## How to test

Verify the project builds

Dependency bumps can resume

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/setup-java@v3.11.0
         with:
           java-version: '11'
-          distribution: 'adopt'
+          distribution: 'corretto'
           cache: 'sbt'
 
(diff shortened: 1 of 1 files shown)


nicl (approved):

.github/workflows/ci.yml
```diff
@@ -35,10 +35,10 @@ jobs:
         uses: actions/setup-node@v3.6.0
 
       # See https://github.com/actions/setup-java
-      - name: Setup Java 8
+      - name: Setup Java 11
         uses: actions/setup-java@v3.11.0
         with:
-          java-version: '8'
+          java-version: '11'
           distribution: 'adopt'
```
nicl, line 42: Can we change this to 'corretto' at the same time? Not sure why adopt was originally used, but feels like a good chance to align with our best practice here.

PR #534, opened 2023-06-13
chore: Remove custom Scala Steward workflows

akash1810:
## What does this change?
We have three GitHub workflows relating to Scala Steward:
  1. `set-automerge.yml` auto-merging in PRs to the `dependency-updates_1` branch
  2. `tracking-branch.yml` keeping the `dependency-updates_1` branch up to date with `main`
  3. `pr-tracking-branch.yml` running once a month, to open a PR against `main` with all the changes from `dependency-updates_1`

The effect of this is we get one mega-PR each month with all the Scala Steward changes batched together.

This is now redundant following an upstream [update](https://github.com/guardian/scala-steward-public-repos/pull/40) to Scala Steward which natively batches changes.

We no longer need our custom version, so remove the workflow files.


tjsilver (approved):
👍

PR #651, opened 2024-01-09
Bump Amiable to Scala 3 and Play 3

AshCorr:
## What does this change?

Bump Amiable to Scala 3 and Play 3, removing Akka in the process.

## How to test

`sbt test` should run in CI.

Deployed in CODE and everything seems functional

### DIFF
--- a/app/services/Agents.scala
+++ b/app/services/Agents.scala
@@ -1,20 +1,19 @@
 package services
 
-import akka.actor.ActorSystem
-import akka.agent.Agent
-import config.AmiableConfigProvider
+import org.apache.pekko.actor.ActorSystem
+import config.{AMIableConfig, AmiableConfigProvider}
 
 import javax.inject.Inject
 import metrics.{CloudWatch, CloudWatchMetrics}
-import models._
+import models.*
 import org.joda.time.DateTime
 import play.api.inject.ApplicationLifecycle
 import play.api.{Environment, Logging, Mode}
 import prism.{Prism, PrismLogic}
-import rx.lang.scala.Observable
 import utils.Percentiles
 
-import scala.concurrent.duration._
+import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.duration.*
 import scala.concurrent.{ExecutionContext, Future}
 
 case class OldInstanceAccountHistory(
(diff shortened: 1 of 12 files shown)


app/services/Agents.scala
```diff
@@ -32,23 +31,23 @@ class Agents @Inject() (
...
+      : AtomicReference[List[(DateTime, Double)]] = AtomicReference(Nil)
   private val amisAgePercentile50thHistoryAgent
-      : Agent[List[(DateTime, Double)]] = Agent(Nil)
+      : AtomicReference[List[(DateTime, Double)]] = AtomicReference(Nil)
   private val amisAgePercentile75thHistoryAgent
-      : Agent[List[(DateTime, Double)]] = Agent(Nil)
+      : AtomicReference[List[(DateTime, Double)]] = AtomicReference(Nil)
   private val oldInstanceCountByAccountAgent
-      : Agent[List[OldInstanceAccountHistory]] = Agent(Nil)
+      : AtomicReference[List[OldInstanceAccountHistory]] = AtomicReference(Nil)
```
AshCorr, line 50: The `Agent` class no longer exists in Pekko (it was removed in Akka 2.5). As I understand this class was meant to be used to ensure whatever data the agent wraps is accessed in a thread-safe & lock-free manner.

Looking at how Amiable uses `Agent` it seems it doesn't actually need to be thread safe, Only one thread writes to each Agent and it overrides the value each time.

app/services/Metrics.scala
```diff
@@ -1,26 +1,32 @@
...
+    system: ActorSystem
 ) {
+  implicit val ec: ExecutionContext = system.dispatcher
+
   if (shouldCreateMetrics) {
 
-    val subscription = Observable
-      .interval(initialDelay = 10.seconds, period = 6.hours)
-      .subscribe { _ =>
+    val subscription = system.scheduler.scheduleAtFixedRate(
```
adamnfish, line 25: One gotcha here is to check what happens if you shut the laptop for a few hours (e.g. an evening) and then open it again. Does it try and catch up all the missed invocations, causing a massive pile of API requests all at once? Or does it pick up from where it is when it wakes? The old Observable approach was safe on this point, I don't know off the top of my head how a Pekko actor schedule will work!

I don't think this is a massive deal, since no one really works on this app anymore* so this is more a point of interest than a request for us to dive into it now 😄

test/prism/JsonUtilsTest.scala
```diff
@@ -27,20 +27,20 @@ class JsonUtilsTest
...
-        'creationDate(Some(new DateTime(2016, 2, 11, 0, 0))),
-        'state("available"),
-        'architecture("x86_64"),
-        'ownerId("0123456789"),
-        'virtualizationType("hvm"),
-        'hypervisor("xen"),
-        'rootDeviceType("ebs"),
-        'sriovNetSupport(Some("simple")),
-        'upgrade(None)
+        Symbol("arn")("arn:aws:ec2:region::image/ami-example"),
```
adamnfish, line 30: Nice one. We should migrate the API off Symbols at some point, but this is a good fix in the meantime 👍

build.sbt
```diff
@@ -58,22 +51,19 @@ libraryDependencies ++= Seq(
   jdbc,
   ehcache,
   ws,
-  "com.typesafe.akka" %% "akka-agent" % "2.5.32",
-  "io.reactivex" %% "rxscala" % "0.27.0",
   "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsSdkVersion,
   "com.amazonaws" % "aws-java-sdk-ses" % awsSdkVersion,
   "com.google.code.gson" % "gson" % "2.10.1",
-  "com.gu.play-googleauth" % "play-v28_2.13" % "2.4.0",
+  "com.gu.play-googleauth" %% "play-v30" % "3.0.6",
```
adamnfish, line 57: Great fix. I'd love to know the story behind why this line was the way it was! 🤪

adamnfish (approved):
Assuming it starts and runs, LGTM. Nice one.

app/services/Agents.scala
AshCorr: We use `scheduleWithFixedDelay` to start the thread that writes to these `Agents`/`AtomicReferences`, as I understand the Pekko scheduler wont start a new thread until the old one has finished, so it _should_ only be getting written to by one thread at a time.

I guess the risk would be that the writer thread doesn't write everything in one go. For example if the writer thread pushes one value at a time to an array and that array is accessed by a reader whilst the writer is still working, the reader could get incomplete data, however in this case we only use `.set` operations replacing one complete array with another complete array.

app/services/Metrics.scala
AshCorr: It should only start one instance!

The way `scheduleAtFixedRate` works under the hood is that after the Runnable completes it will schedule the next run.

I would have kept Observabale, but the library its from (rxscala) appears to be unmaintained now and doesn't have a Scala 3 version.

PR #652, opened 2024-01-10
Remove Amiable internal Google Auth

AshCorr:
## What does this change?

Removes all Google Authenticator code from Amiable. We now authenticate at the LB making the double login a bit superflous. This also makes running Amiable locally significantly easier.

One thing to check before merging this PR is whether both the LB and amiable were configured to restrict access to the same Google Groups. We don't want to widen the scope of who can access Amiable by accident.

## How to test

Run all tests & Deploy to CODE.


AshCorr:
As per https://github.com/guardian/security-recommendations/blob/main/recommendations/securing-staff-services.md#authentication--authorisation-at-the-application we should still maintain some level of authentication in the application itself.

PR #662, opened 2024-02-23
send email for giant instances with old ami

marjisound:
## What does this change?

In investigations account, for one of our services Giant, we are not following the same pattern for stage & stack & app as it's done for all other instances. 

For Giant, the pattern is: 
- Stage: `rex`
- Stack: `pfi-playground` | `pfi-giant`
- App: `pfi` | `pfi-worker` | `elasticsearch` | `neo4j`

because of this, we are not receiving email notifications for giant instances with old ami.


philmcmahon (approved):

akash1810 (approved):
I think it would be better to be generic here, and we default to inspecting _all_ instances. If there's a specific reason for an instance to be hidden from this alert (e.g. it holds state) then we express that via a tag.

That is, make this an opt-out feature, rather than opt-in.

That said, this is a great improvement to the status quo 👍🏽.

@guardian/devx-operations and @guardian/devx-security have (future) plans to deprecate AMIable. Maybe we can add this to our thinking?

PR #663, opened 2024-02-28
Automatically determine obsolete Ubuntu distros

adamnfish:
## What does this change?

Replaces hard-coded list of supported Ubuntu versions with logic to work this out, based on the current date. This is done using canonical's appraoch, which is that LTS releases are April in even numbered years, with 5 year support. All other releases have just 9 months support.

We change the logic itself, update the wiring in the template, and add a few more tests for the date logic.

Getting the current date from the template (to pass to the helper function) is not a great pattern, but this is consistent with the current approach in these templates.

## How to test

I've added tests, and we can glance at the UI to make sure that it is correctly identifying obsolete images, if any exist.

It should be clearer that teams need to update their base image when it drops off support.

If canonical ever change their approach, this logic will no longer be accurate. This hasn't happened yet, and likely Amiable won't be in use if that comes up.

### DIFF
--- a/app/prism/Recommendations.scala
+++ b/app/prism/Recommendations.scala
@@ -1,6 +1,9 @@
 package prism
 
 import models.AMI
+import org.joda.time.DateTime
+
+import scala.util.Try
 
 object Recommendations {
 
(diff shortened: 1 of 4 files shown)


app/prism/Recommendations.scala
```diff
@@ -65,21 +68,20 @@ object Recommendations {
     }) getOrElse false
   }
 
-  def isObsoleteUbuntu(ubuntuAmi: AMI): Boolean = {
-    val DistExtractor = ".*/ubuntu-(\\w+?)-(\\d{2}.\\d{2})-.*".r
+  def isObsoleteUbuntu(ubuntuAmi: AMI, now: DateTime): Boolean = {
+    val DistExtractor = ".*/ubuntu-(\\w+?)-(\\d{2}).(\\d{2})-.*".r
```
NovemberTang, line 72: Is there a way to do this without using regex? or maybe we could add a comment explaining what kind of pattern this is looking for

AshCorr (approved):

app/prism/Recommendations.scala
```diff
@@ -65,21 +68,20 @@ object Recommendations {
...
+      case DistExtractor(distName, distYear, distMonth) =>
+        Try {
+          val year = distYear.toInt
+          val month = distMonth.toInt
+          val releaseDate = new DateTime(2000 + year, month, 1, 0, 0)
+          if (year % 2 == 0 && distMonth == "04") {
+            releaseDate.plusYears(5).isBefore(now)
+          } else {
+            releaseDate.plusMonths(9).isBefore(now)
+          }
```
AshCorr, line 83: A slight blurb on why this logic is needed might be helpful for future code divers!

PR #708, opened 2024-08-15
remove nvm from cdk ci script

tjsilver:
# What does this change?

While debugging a failed version bump we discovered that nvm appears to no longer be available on the Github runner. This meant that CI was broken.

## How to test

CI should pass.

### DIFF
--- a/cdk/script/ci
+++ b/cdk/script/ci
@@ -2,23 +2,6 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


adamnfish (approved):
We looked at this together, and discovered that this was what was breaking the build for another recent PR.

LGTM!

PR #709, opened 2024-08-15
fix: bump jackson to remediate vulnerability

tjsilver:
## What does this change?

This issue is similar to that in [security-hq](https://github.com/guardian/security-hq/pull/1149) and should remediate the same issue by bumping jackson to v17.


adamnfish (approved):
LGTM

As ever, we'll have to be careful to watch out for runtime errors from incompatible versions, but we're pretty sure this will be fine.

PR #738, opened 2024-12-04
Ensure sbt is properly installed in GitHub actions

adamnfish:
## What does this change?

Using the Guardian-provided [setup-scala](https://github.com/guardian/setup-scala) workflow allows us to configure and install Java and sbt at the same time. It takes the Java version out of the .tool-versions file, so this file is also added with a (currently) up-to-date Java version.

## How to test

We should see the workflows complete successfully now.

We'll avoid these workflows breaking when GitHub starts to roll-out the new base image from 5th Dec.

We're specifying a needlessly specific Java version in the .tool-versions file, to appease `asdf`. We will separately migrate to just the major version by adopting [mise](https://mise.jdx.dev/), but this is something to change globally when we are ready.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,13 +33,9 @@ jobs:
           cache: 'npm'
           cache-dependency-path: 'cdk/package-lock.json'
 
-      # See https://github.com/actions/setup-java
-      - name: Setup Java 11
-        uses: actions/setup-java@v4.5.0
-        with:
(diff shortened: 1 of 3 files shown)


jorgeazevedo (approved):

akash1810:
https://github.com/guardian/amiable/pull/740 is failing CI with "sbt not found". Merging this to resolve.

akash1810 (approved):

PR #767, opened 2025-04-07
Fix CODEOWNERS file

kelvin-chappell:
If a file group is repeated, only the last line applies.  So in this case, only the security team were notified of a PR.

See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax.


tjsilver (approved):
Good spot!

PR #799, opened 2025-08-18
AWS Java SDK v2 migration

adamnfish-gu:
## What does this change?

Updates to the AWS v2 Java SDK. The v1 SDK is EoL at the end of the year, so it's important we make this change in the near future.

We should make sure everything still works exactly as it did before.

It has been tested on CODE, including triggering an email send.

The tricky thing to test will be the SES integration that sends email digests. I'm open to advice on how to test this, or whether we should release it and keep a careful eye out.


app/services/notification/AmazonSimpleEmailServiceAsyncFactory.scala
```diff
@@ -1,28 +1,41 @@
...
-    Region.getRegion(Regions.EU_WEST_1)
-  )
+    val profileProvider = ProfileCredentialsProvider
+      .builder()
+      .profileName("deployTools")
+      .build()
+
+    val instanceProvider = InstanceProfileCredentialsProvider
+      .builder()
+      .build()
```
kelvin-chappell, line 24: Could these be shared with the Cloudwatch client?

app/services/notification/AWSMailClient.scala
```diff
@@ -1,25 +1,35 @@
...
-    implicit exec: ExecutionContext
+class AWSMailClient @Inject() (amazonMailClient: SesAsyncClient)(implicit
+    exec: ExecutionContext
 ) extends Logging {
 
   def send(toAddress: String, request: SendEmailRequest): Attempt[String] = {
 
-    val messageId =
-      awsToScala(amazonMailClient.sendEmailAsync)(request).map(_.getMessageId)
+    val updatedRequest = request
```
kelvin-chappell, line 19:
```suggestion
    val requestWithDestination = request
```

kelvin-chappell (approved):
✨

PR #890, opened 2026-03-09
Add dev container support

aug24:
## What does this change?
Adds devcontainer support.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214110351187533

### DIFF
--- a/.devcontainer/.gitignore
+++ b/.devcontainer/.gitignore
@@ -1,0 +1,2 @@
+# User-specific devcontainer directory with merged personal preferences
+user/
--- a/.devcontainer/devenv.yaml
+++ b/.devcontainer/devenv.yaml
@@ -25,8 +25,8 @@ forwardPorts:
   -  9101
 
 # Optional: Mount directories from host to container
-# mounts:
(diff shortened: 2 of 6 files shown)


adamnfish-gu (commented):
We'll need the port forwards and configuration mounts for this project to be properly onboarded.

As a rule of thumb, the project should be editable in the IDE and it should run locally. Here, the latter will need its ports forwarded and the local config directory mounted (I think it can be read-only).

adamnfish-gu (approved):

